PR #24, opened 2023-11-28
[WIP] Proposal to cleanup requirements

kachayev:
Technically, neither `pandas` nor `matplotlib` are required when using skada as a library. It might be useful for running demos but for those we should have either docs or suggestions in the demos itself.


rflamary (commented):
I'm OK with this.

maybe we should have 3 requirement file, one for documentatoion (in folder doc), one called requiremet with minimum requirement (thta is not even torch) and one with requirement_all with all requireme,t for 100% coverage of teh functions. This would also lead to tsts that can adapt to installed modules

rflamary:
Closing this issue I cleanup up teh requiements

PR #27, opened 2023-12-02
Rename `pack_flatten` to `pack_for_lodo`

### LINKED ISSUE #5: Rename dataset method `pack_flatten` to `pack_for_lodo`
See `skada.datasets.DomainAwareDataset`. The method is quite edgy, as of now it's sole purpose to pack the dataset for a `LeaveOneDomainOut` splitter.

Rename and write docstring with the proper explanation of what this method is doing (and when it should be used).

kachayev:
Change the name of the method to be precise that the out is only useful when doing Leave-One-Domain-Out.

Additional docstrings are provided as well.

Closes #5.

### DIFF
--- a/skada/datasets/_base.py
+++ b/skada/datasets/_base.py
@@ -25,6 +25,11 @@ DomainDataType = Union[
     Tuple[np.ndarray,],
 ]
 
+PackedDatasetType = Union[
+    Bunch,
+    Tuple[np.ndarray, np.ndarray, np.ndarray]
+]
+
 
 def get_data_home(data_home: Union[str, os.PathLike, None]) -> str:
     """Return the path of the `skada` data folder.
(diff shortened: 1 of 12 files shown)


skada/datasets/_base.py
```diff
@@ -253,15 +259,49 @@ def pack_for_test(
...
             return_X_y=return_X_y,
             train=False,
         )
 
-    def pack_flatten(self, return_X_y: bool = True):
+    def pack_for_lodo(self, return_X_y: bool = True) -> PackedDatasetType:
+        """Packages all domains in a format compatible with the Leave-One-Domain-Out
+        cross-validator (refer to :class:`~skada.model_selection.LeaveOneDomainOut` for
+        more details). To enable the splitter's dynamic assignment of source and target
+        domains, data from each domain is included in the output twice.
```
rflamary, line 274:
```suggestion
        domains, data from each domain is included in the output twice (as a source and a target).
```

skada/datasets/_base.py
```diff
@@ -253,15 +259,49 @@ def pack_for_test(
...
+
+        Exercise caution when using this output for purposes other than its intended
+        use, as this could lead to incorrect results.
```
rflamary, line 277:
```suggestion
        use, as this could lead to incorrect results and data leakage.
```

tgnassou:
Do we really need the for in the name ? And same question for train and test ? I think `pack_train`, `pack_test` and `pack_lodo` are enough ?

tgnassou (approved):

PR #42, opened 2024-01-09
[DOC] Corrections README and string for DomainAware dataset

YanisLalou:
- Replace TCLAdapter by CORALAdapter
- Change typo error


rflamary:
One day we will add something better than CORAL Adapter ;)

kachayev:
@YanisLalou It seems you have 2 changes in the same PR: one about the README, and one about string representation for datasets. Those are better be done separately, so it's just easier to discuss them.

Why do we want to switch to CORAL?

YanisLalou:
@kachayev We couldn't locate the TCLAdapter class in the Git repository, so we opted for a random already implemented adapter.

kachayev:
@YanisLalou Oh, right. We use the full name in the code. So it's `TransferComponentAnalysisAdapter`.

rflamary (approved):
I'm OK with this PR next time we splitthe PR for difefernt things indeed : justr corect the PR name befor merging to reflct thwhat it does

Issue #25, opened 2023-11-30
Rename Bunch keys to avoid confusion


kachayev:
Rename "data" to "X" and "targets" to "y", to avoid confusion with "target domains".

PR #44, opened 2024-01-09
Rename Bunch keys

YanisLalou:
#25 
- Rename Bunch keys in the `DomainAwareDataset` def

### DIFF
--- a/skada/datasets/_base.py
+++ b/skada/datasets/_base.py
@@ -232,8 +232,8 @@ class DomainAwareDataset:
         target = np.concatenate(ys)
         sample_domain = np.concatenate(sample_domains)
         return (data, target, sample_domain) if return_X_y else Bunch(
-            data=data,
-            target=target,
+            X=data,
+            y=target,
             sample_domain=sample_domain,
             domain_names=domain_labels,
         )
--- a/skada/datasets/_base.py
+++ b/skada/datasets/_base.py
@@ -292,10 +292,10 @@ class DomainAwareDataset:
         data : :class:`~sklearn.utils.Bunch`
             Dictionary-like object, with the following attributes.
 
-            data: np.ndarray
+            X: ndarray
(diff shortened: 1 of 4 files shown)


skada/datasets/_base.py
```diff
@@ -232,8 +232,8 @@ def pack(
         target = np.concatenate(ys)
         sample_domain = np.concatenate(sample_domains)
         return (data, target, sample_domain) if return_X_y else Bunch(
```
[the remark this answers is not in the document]
YanisLalou: To avoid confusion with the X and y in the for loop, I reuse the names Xs and ys instead of data and target respectively

kachayev:
It seems those we mentioned in the docs as well (in the README file).

kachayev:
Plus sample generator functions have documentation (not sure if they implement packing locally but worth checking out).

kachayev:
The documentation for `pack_lodo` method of the base dataset has the same set of keys as well.

kachayev:
Same for Office31 datasets.

kachayev:
It seems the `_load_office31` loader for the Office31 dataset still returns keys with old names.

skada/datasets/_office.py
```diff
@@ -470,4 +470,7 @@ def _load_office31(
             data.append(content.astype(loader_spec.dtype))
     files['data'] = np.vstack(data)
```
rflamary, line 471: just set X and y here no need to add stuff and pop it

YanisLalou: We can't. The `files` object comes from the `load_files` function, which is a sklearn function. Thus the keys 'data' and 'target' exists from the beginning.
Thus at some point we would have to remove these keys to not have duplicates with the keys 'X' and 'y'.
Thus the pop methods

kachayev (approved):

PR #45, opened 2024-01-10
[MRG] DA deep methods with new API

tgnassou:
PR related to Issue #1 . It's the first draft of the feature method reformatting. I changed all the logic behind these methods without changing skorch. I implemented Custom criterion, module, and dataloaders for `X, y, sample_domain`.  I added an example for DeepCoral method, let me know if you have suggestions for the idea, the implementation is not really fixed.

PR #47, opened 2024-01-10
[TO_REVIEW] Make DeepEmbeddedValidation scorer to work with the new API

### LINKED ISSUE #2: Make `DeepEmbeddedValidation` scorer to work with the new API
As of now `skada.metrics.DeepEmbeddedValidation` scorer doesn't work with a new API. Not sure why. The test suite `tests/test_scorer.py` has it commented out because of this.

YanisLalou:
Fixes #2 
Paper link (also page 5): https://proceedings.mlr.press/v97/you19a/you19a.pdf

- To compute the features we find the last layer of the pipeline with a transform method or we use the .predict_features() function if its a deep model


skada/metrics.py
```diff
@@ -370,3 +396,24 @@ def _score(self, estimator, X, y, sample_domain=None, **kwargs):
         var_w = np.var(weights, ddof=1)
         eta = -cov / var_w
         return self._sign * (np.mean(weighted_error) + eta * np.mean(weights) - eta)
+    
+
+    def cross_entropy_loss(self, y_true, y_pred, epsilon=1e-15):
```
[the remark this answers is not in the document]
YanisLalou: In our case yes because we are computing the CE loss for each sample individually.
The **sklearn** implementation throws an error if at least one of each possible label is not present in `y_pred`.
`Error: ValueError: The number of classes in labels is different from that in y_pred.`

rflamary:
Question: do we merge this PR so that it works with a pipeline or do we wait for the Feature deep methods of @tgnassou in PR #45 ?

skada/metrics.py
```diff
@@ -328,39 +328,65 @@ def __init__(
...
         rng = check_random_state(self.random_state)
         X_train, X_val, _, y_val, _, sample_domain_val = train_test_split(
             X[source_idx], y[source_idx], sample_domain[source_idx],
             test_size=0.33, random_state=rng
         )
-        features_train = estimator.get_features(X_train)
-        features_val = estimator.get_features(X_val)
-        features_target = estimator.get_features(X[~source_idx])
 
+        features_train = estimator[:index_transformer+1].transform(X_train)
```
tgnassou, line 380: You want to have a transform method, but this DEV scorer is also made (and mainly) to work with deep learning. I didn't finish the PR but there will be no transform function for these feature methods.

skada/metrics.py
```diff
@@ -328,39 +328,65 @@ def __init__(
...
+        features_target = estimator[:index_transformer+1].transform(X[~source_idx])
+        
         self._fit_adapt(features_train, features_target)
-        N, N_target = len(features_train), len(features_target)
-        predictions = self.domain_classifier_.predict_proba(features_val)
-        weights = N / N_target * predictions[:, :1] / predictions[:, 1:]
-
-        y_pred = estimator.predict_proba(X_val, sample_domain=sample_domain_val)
+        N_train, N_target = len(features_train), len(features_target)
+        predictions = self.domain_classifier_.predict_proba(features_val.get('X'))
```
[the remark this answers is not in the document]
tgnassou: I still think the name can be better here, `predictions` is not clear.

tgnassou:
This scorer will not work with the new API of features method. Most of the changes will have to be changed again afterward, so maybe we can wait a few weeks before merging. wdyt @YanisLalou ?

skada/deep/base.py
```diff
@@ -386,6 +387,45 @@ def predict(self, X, sample_domain=None, **predict_params):
...
+            The input data. If a dict, it should contain a key 'X' with the
+            input data and a key 'sample_domain' with the domain of each
+            sample.
+            If X is a dataset, the dataset should return a dict.
+        sample_domain : torch tensor
+            The domain of each sample.
+            Could be None since the sample are not used in predict.
+        """
+        if isinstance(X, dict):
+            if "X" not in X.keys():
```
tgnassou, line 405: You can easily coverage these lines by duplicating the `test_base` for `predict_proba`

skada/deep/modules.py
```diff
@@ -29,6 +30,33 @@ def forward(
         return X
 
 
+class ToyModuleClassification2D(torch.nn.Module):
+    """XXX add docstring here."""
```
tgnassou, line 34: maybe we can just put a parameter `softmax` instead of creating a new class

skada/metrics.py
```diff
@@ -365,13 +366,36 @@ def _fit_adapt(self, features, features_target):
...
         return self
 
     def _score(self, estimator, X, y, sample_domain=None, **kwargs):
+        if not hasattr(estimator, "predict_proba"):
+            raise AttributeError(
+                "The estimator passed should have a 'predict_proba' method. "
+                f"The estimator {estimator!r} does not."
+            )
+
+        if not isinstance(estimator, BaseEstimator):
```
tgnassou, line 382: Is this test very trustful? A non-deep model can still be a `BaseEstimator` no?

skada/deep/tests/test_deep_scorer.py
```diff
@@ -0,0 +1,74 @@
...
+)
+from skada.deep.modules import ToyModuleClassification2D
+from skada.metrics import (
+    DeepEmbeddedValidation,
+    PredictionEntropyScorer,
+    SoftNeighborhoodDensity,
+)
+
+
+class TestLoss(BaseDALoss):
```
tgnassou, line 24: If we use it in different tests maybe we can create a '_TestLoss` class somewhere

skada/metrics.py
YanisLalou: Yes here there's 2 possibilities:
Either the estimator inherits from BaseEstimator or inherits from Nn.module.
But I wonder how it'll behave if we have a nn.module inside a sklearn.pipeline 🤔

skada/deep/tests/test_deep_divergence.py
```diff
@@ -1,20 +1,25 @@
...
 from skada.datasets import make_shifted_datasets
 from skada.deep import DAN, DeepCoral
 from skada.deep.modules import ToyModule2D
 
 
 def test_deepcoral():
+    # Set seed for reproducibility
+    np.random.seed(42)
+    torch.manual_seed(42)
+
```
antoinecollas, line 22: There is the fixture `set_seed` in `conftest.py` that set the seed for every function. Can you add `torch.manual_seed(42)` to this fixture?

skada/metrics.py
```diff
@@ -381,22 +381,21 @@ def _score(self, estimator, X, y, sample_domain=None, **kwargs):
...
         else:
             # We need to find the last layer of the pipeline with a transform method
             pipeline_steps = list(enumerate(estimator.named_steps.items()))
 
             for index_transformer, (_, step_process) in reversed(pipeline_steps):
                 if hasattr(step_process, "transform"):
                     transformer = estimator[: index_transformer + 1].transform
+                    has_transform_method = True
                     break  # Stop after the first occurrence if there are multiple
-            else:
```
tgnassou: Maybe it could be better to define identity function:
`def identity(x): return x` and if there is no transform method, you can put `transformer = identity`. It will be clearer I think

Issue #12, opened 2023-11-21
`__str__` and `__repr__` for `DomainAwareDataset`


kachayev:
We should try to follow common guideline for sklearn dataset, though a few challenges arise with trying to do so. Proposals are most certainly welcome!

kachayev:
This one was closed in #49

PR #49, opened 2024-01-11
DomainAwareDataset str repr edge case handling

YanisLalou:
Issue #12 
- Handling the str output of a DomainAwareDataset if the number of domin_names_ is too large
- Print the number of datapoints in the dataset

### DIFF
--- a/skada/datasets/_base.py
+++ b/skada/datasets/_base.py
@@ -322,7 +322,7 @@ class DomainAwareDataset:
         body.append(f"Total size: {sum(len(tup[0]) for tup in self.domains_)}")
         output = "\n".join([head] + body)
         return output
-    
+
     def _get_domain_representation(self, max_domains=5, max_length=50):
         domain_names = list(self.domain_names_.keys())
 
--- a/skada/datasets/_base.py
+++ b/skada/datasets/_base.py
@@ -314,10 +314,30 @@ class DomainAwareDataset:
         )
 
     def __str__(self) -> str:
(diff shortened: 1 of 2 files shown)


YanisLalou:
The goal is to have a similar string representation as torchvision for datasets:
`Dataset CIFAR10
    Number of datapoints: 50000
    Root location: datasets/CIFAR10
    Split: Train`

However rn in the `DomainAwareDataset` object we don't have the Dataset name + Root location

skada/datasets/_base.py
```diff
@@ -314,10 +314,31 @@ def pack_lodo(self, return_X_y: bool = True) -> PackedDatasetType:
...
 
     def __repr__(self) -> str:
-        return self.__str__()
+        head = self.__str__()
+        body = [f"Number of domains: {len(self.domains_)}"]
+        body.append(f"Total size: {sum(len(tup[0]) for tup in self.domains_)}")
+        output = "\n".join([head] + body)
+        return output
+    
+    def get_domain_representation(self, max_domains=5, max_length=50):
```
kachayev, line 326: I suggest renaming the function to `_get_domain_representation` to emphasize its role as an internal function. This change will help clarify that it should not be considered part of the class's public API.

rflamary (commented):
This seems OK but you need to do some testing (add at leats one test) with different types of datasets and check that nothing breaks

kachayev:
@YanisLalou Flake8 is not happy here as well.

Issue #28, opened 2023-12-02
Unwrap expliticly given selector before generating the name for the pipeline


kachayev:
`make_da_pipeline` function right now gives an option to provide as a step estimator already wrapped into one of the selectors. When we use `_name_estimators`, it generates rather weird name. Better approach would be to unwrap estimator from the selector before naming it.

kachayev:
Make sure #29 is merged before working on this issue.

YanisLalou:
Can you give examples of "weird names"?

In this example:
```
{'otmappingadapter': Shared(base_estimator=OTMappingAdapter(), max_iter=100000, metric='sqeuclidean',
        norm=None),
 'perdomain-1': PerDomain(base_estimator=PCA(), copy=True, iterated_power='auto',
           n_components=None, n_oversamples=10,
           power_iteration_normalizer='auto', random_state=None,
           svd_solver='auto', tol=0.0, whiten=False),
 'perdomain-2': PerDomain(C=1.0, base_estimator=LogisticRegression(), class_weight=None,
           dual=False, fit_intercept=True, intercept_scaling=1, l1_ratio=None,
           max_iter=100, multi_class='auto', n_jobs=None, penalty='l2',
           random_state=None, solver='lbfgs', tol=0.0001, verbose=0,
           warm_start=False)}
           
```
           
Do we want to have instead of  `perdomain-1` and ` perdomain-2,` `pca` and `log_reg` respectively ?

rflamary:
perdomain_pca maybe?

kachayev:
Maybe we can do `perdomain_*` for `PerDomain` selector but remove `shared_*` for `Shared`? I guess, if you put `PerDomain` explicitly, you know what you are doing and what to expect. But `Shared` is added "automatically" (most likely), so you won't know how to refer to your estimator in the pipeline.

kachayev:
Closed in #51

PR #51, opened 2024-01-11
[WIP] Unwrap expliticly given selector before generating the name for the pipeline

YanisLalou:
Issue #28 

- Raise TypeError if no steps are given when `make_da_pipeline` is called

### DIFF
--- a/skada/_pipeline.py
+++ b/skada/_pipeline.py
@@ -79,6 +79,9 @@ def make_da_pipeline(
     """
     # note that we generate names before wrapping estimators into the selector
     # xxx(okachaiev): unwrap from the selector when passed explicitly
+    if not steps:
+        raise TypeError("Missing 1 required positional argument: 'steps'")
+    
     names = [step[0] if isinstance(step, tuple) else None for step in steps]
     estimators = [step[1] if isinstance(step, tuple) else step for step in steps]
     steps = _wrap_with_selectors(_name_estimators(estimators), default_selector)


rflamary (commented):
This lloks fine but you need to do a test for different configuration (look at codecoiv reports for checking that all if/else in name_esimator is tested and that it returns the proper names)

YanisLalou:
No a fan on how we test the `if isinstance(estimator, str)` condition rn.
Don't see how this condition is satisfied in a real use case.

kachayev:
> No a fan on how we test the `if isinstance(estimator, str)` condition rn. Don't see how this condition is satisfied in a real use case.

The default value for `default_selector` is `'shared'`, this should satisfy the condition.

YanisLalou:
We name the estimators before applying the `default_selector`. So we don't even go through the function `_name_estimators()` a 2nd time. Thus it won't satisfy the condition.

Currently, it is not an issue as we have mutually decided that the default selector is always set to `shared`, and we have also agreed not to display the `shared_` preposition before each estimator name. 
However, if we intend to implement a modifiable `default_selector`, such as `Per_domain`, the estimator will not be named correctly in our current implementation.
Indeed in the latter case, we should apply the `_wrap_with_selectors` function before naming the estimators.

kachayev:
Surely parametrizable `default_selector` functionality should remain working after changes made to the naming convention, if that was the question.

YanisLalou:
What I meant is that right now by applying the `_name_estimators` function before the `_wrap_with_selectors` one, we don't comply with the naming conventions that we agreed to.
So I'll modify that

kachayev (approved):

Issue #17, opened 2023-11-21
Make sure all API methods accept `sample_domain` as `None`


kachayev:
In this case the method has to derive domain labels using `y` masking. If masking is not done or, in a way, ambiguous, we should throw `ValueError` to report inconsistent inputs.

Typically done with `allow_auto_sample_domain=True` param for `check_*_domain` utils. I'm not sure this is implemented consistently throughout all methods. Also, test cases to cover this are absolutely necessary.

YanisLalou:
By default `allow_auto_sample_domain` is set to false in the `check_X_y_domain` function. Doesn't it make more sense to set it to true by default?

kachayev:
Yes, I agree that would be a more sensible default option.

PR #53, opened 2024-01-12
[MRG] Make sure all API methods accept sample_domain as None

YanisLalou:
Issue #17

- Change default value for `allow_auto_sample_domain` in `check_X_y_domain`
- check if the masking of y is well done in `check_X_y_domain`


skada/_utils.py
```diff
@@ -59,11 +59,18 @@ def check_X_y_domain(
     X = check_array(X, input_name='X', allow_nd=allow_nd)
     y = check_array(y, force_all_finite=True, ensure_2d=False, input_name='y')
     check_consistent_length(X, y)
-    if sample_domain is None and allow_auto_sample_domain:
+    if sample_domain is None and not allow_auto_sample_domain:
```
kachayev, line 62: We need to make sure that all of those clauses are properly covered with unit tests. I guess just updating `check_X_y_domain_exceptions` with a bunch of correct and incorrect inputs would do.

skada/_utils.py
```diff
@@ -59,11 +59,18 @@ def check_X_y_domain(
...
+        raise ValueError("Either 'sample_domain' or 'allow_auto_sample_domain' "
+                         "should be set")
+    elif sample_domain is None and allow_auto_sample_domain:
+        y_type = _check_y_masking(y)
         sample_domain = np.ones_like(y)
         # labels masked with -1 are recognized as targets,
         # the rest is treated as a source
-        sample_domain[y == -1] = -2
+        if y_type == 'classification':
+            sample_domain[y == -1] = -2
```
kachayev, line 71: I always try to avoid using 'magic' numbers in the code, as they can be extremely difficult to understand and modify later on. Let's create a constant named `_DEFAULT_TARGET_DOMAIN_LABEL` = -2 (or something like this) in the current module namespace.

skada/_utils.py
```diff
@@ -59,11 +59,18 @@ def check_X_y_domain(
...
+        else:
+            sample_domain[np.isnan(y)] = -2
```
kachayev, line 73: Same constant here.

skada/_utils.py
```diff
@@ -114,13 +121,17 @@ def check_X_domain(
     return_indices: bool = False,
     # xxx(okachaiev): most likely this needs to be removed as it doesn't fit new API
     return_joint: bool = True,
-    allow_auto_sample_domain: bool = False,
+    allow_auto_sample_domain: bool = True,
```
kachayev, line 124: Let's put a docstring for this function.

skada/_utils.py
```diff
@@ -40,15 +40,15 @@ def _estimate_covariance(X, shrinkage):
...
+    sample_domain = None,
     allow_source: bool = True,
     allow_multi_source: bool = True,
     allow_target: bool = True,
     allow_multi_target: bool = True,
     return_indices: bool = False,
     # xxx(okachaiev): most likely this needs to be removed as it doesn't fit new API
     return_joint: bool = False,
-    allow_auto_sample_domain: bool = False,
+    allow_auto_sample_domain: bool = True,
```
kachayev, line 51: Let's put a docstring for this function as well.

kachayev:
@YanisLalou I can't seem to push the update from my local branch here (not sure why, github previously allowed me to do so). Would you please check `flake8` locally and fix all of its complaints?

skada/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    with pytest.raises(ValueError):
+        _check_y_masking(y_wrongfuly_masked_1)
+
+    with pytest.raises(ValueError):
+        _check_y_masking(y_wrongfuly_masked_2)
+
+
+def test_check_y_masking_regression():
+    y_properly_masked = np.array([np.nan, 1, 2.5, -1, np.nan, 0, -1.5])
+    y_wrongfuly_masked = np.array([-1, -2, 2.5, -1, 2, 0, 1])
```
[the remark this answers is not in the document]
YanisLalou: I don't get it, isn't it possible to have masked arrays for regression tasks?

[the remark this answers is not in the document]
kachayev: It's possible to have masks for regression using `nan`s. The variable that's called `y_wrongfuly_masked` doesn't have a 'wrongly' masked array, it does have 'non masked' array. That's why I was confused about the name.

skada/_utils.py
```diff
@@ -40,30 +42,86 @@ def _estimate_covariance(X, shrinkage):
...
+    elif sample_domain is None and allow_auto_sample_domain:
+        y_type = _check_y_masking(y)
         sample_domain = np.ones_like(y)
-        # labels masked with -1 are recognized as targets,
+        # labels masked with -1 (for classification) are recognized as targets,
+        # labels masked with nan (for regression) are recognized as targets,
         # the rest is treated as a source
-        sample_domain[y == -1] = -2
+        if y_type == 'classification':
+            sample_domain[y == -1] = _DEFAULT_TARGET_DOMAIN_LABEL
```
kachayev, line 122: Let's move this constant '1' to a named variable as well (same as it was done for `_DEFAULT_TARGET_DOMAIN_LABEL`)

skada/_utils.py
```diff
@@ -114,13 +172,61 @@ def check_X_domain(
...
-    if sample_domain is None and allow_auto_sample_domain:
+    if sample_domain is None and not allow_auto_sample_domain:
+        raise ValueError("Either 'sample_domain' or 'allow_auto_sample_domain' "
+                         "should be set")
+    elif sample_domain is None and allow_auto_sample_domain:
         # default target domain when sample_domain is not given
         # xxx(okachaiev): I guess this should be -inf instead of a number
-        sample_domain = -2*np.ones(X.shape[0], dtype=np.int32)
+        # The idea is that with no labels we always assume target domain (-1)
+        sample_domain = -1*np.ones(X.shape[0], dtype=np.int32)
```
kachayev, line 229: We put here -1s but we also have `_DEFAULT_TARGET_DOMAIN_LABEL`, which is 2. Should those be the same? 🤔

YanisLalou: I think that they don't should be the same. 
Here we assume that all datapoints are coming from the same target_domain and that there is no source_domain. Thus we need ONE domain value and it needs to have a negative sign. And if we want to follow the convention of assigning domain values first from (1 to n) and then modifying their sign if its a target domain, we should keep here the -1.

kachayev: Good, I'm okay with this. Let's create a named constant for this situation as well.

skada/_utils.py
```diff
@@ -40,30 +51,87 @@ def _estimate_covariance(X, shrinkage):
...
     X,
     y,
-    sample_domain,
+    sample_domain=None,
     allow_source: bool = True,
     allow_multi_source: bool = True,
     allow_target: bool = True,
     allow_multi_target: bool = True,
     return_indices: bool = False,
     # xxx(okachaiev): most likely this needs to be removed as it doesn't fit new API
```
rflamary, line 60: if we shoudl remove something aybe do it while we are working on ythis function?

skada/_utils.py
```diff
@@ -114,13 +182,64 @@ def check_X_domain(
...
+    return_indices : bool, optional (default=False)
+        If True, return only the source indices.
+    return_joint : bool, optional (default=True)
+        If True, return X and sample_domain without separation.
+    allow_auto_sample_domain : bool, optional (default=True)
+        Allow automatic generation of sample_domain if not provided.
+
+    Returns:
+    ----------
+    If return_indices is True:
```
rflamary, line 217: do we erally want the output of the function to change this much? The way I see that this function does 4 different thgigs:
- input validation (is everythin OK), this is always done and raise error if not
- Return info or data
  - indices
  - X_source/X_taregt
  - Return joint (with pitentially updated sample_domain if None)
  
 I think we should factorized the data validztion and have 3 functiosn for the 3 different return values maybe? We can talk about this i slack

rflamary (commented):
A few comments maybe we should discuss that in the chat

skada/_utils.py
kachayev: I bet there's a separate task for this)

YanisLalou:
⬆️ Issue #58

skada/tests/test_utils.py
```diff
@@ -0,0 +1,308 @@
...
+        X, y, sample_domain
+    )
+
+    assert len(X_source) == 2 * n_samples_source, "X_source shape mismatch"
+    assert len(y_source) == 2 * n_samples_source, "y_source shape mismatch"
+    assert len(X_target) == 2 * n_samples_target, "X_target shape mismatch"
+    assert len(y_target) == 2 * n_samples_target, "y_target shape mismatch"
+
+
+def split_source_target_X():
```
kachayev, line 294: Tests should start with `test_`.

skada/utils.py
```diff
@@ -0,0 +1,264 @@
...
+        dtype=np.int32,
+        ensure_2d=False,
+        input_name='sample_domain'
+    )
+
+    source_idx = (sample_domain >= 0)
+    return source_idx
+
+
+def split_source_target_X_y(X, y, sample_domain):
```
kachayev, line 219: Rather than having separate functions for `split_source_target_X_y` and `split_source_target_X`, it might be simpler to have a single function `split_source_target` following the call convention of `train_test_split`: the function returns pair for each input given. Something like:

```python
X_source, X_target = split_source_target(X, sample_domain=sample_domain)
X_source, X_target, y_source, y_target = split_source_target(X, y, sample_domain=sample_domain)
X_source, X_target, y_source, y_target, D_source, D_target  = split_source_target(X, y, D, sample_domain=sample_domain)
# ... etc
```

Also, maybe `source_target_split` (similar to `train_test_split` in sklearn)?

kachayev (changes requested):

skada/utils.py
YanisLalou: I agree. Just, there is already a `source_target_split` function in the _utils.py file. Should I move it to utils.py and modify it so he can handle more than just (X, y) as a split input?

kachayev: If we have a new function available from utils we don't need an old one

skada/tests/test_utils.py
```diff
@@ -0,0 +1,289 @@
...
+    )
+
+    # Test that no ValueError is raised
+    _, _ = source_target_split(X, sample_domain=sample_domain)
+
+    X_source, X_target, y_source, y_target = source_target_split(
+        X, y, sample_domain=sample_domain
+    )
+
+    print(X.shape)
```
kachayev, line 106: I guess we don't need those print statements in the code

skada/tests/test_utils.py
```diff
@@ -0,0 +1,289 @@
...
+            X, sample_domain=random_sample_domain,
+            allow_auto_sample_domain=False, allow_source=allow_source
+        )
+
+    with pytest.raises(
+        ValueError,
+        match=(
+            f"Number of targets provided is {n_targets} "
+            f"and 'allow_target' is set to {allow_target}"
+            )
```
kachayev, line 243: Does flake8 allow this type of indentation?

skada/utils.py
```diff
@@ -0,0 +1,247 @@
...
+
+def check_X_y_domain(
+    X,
+    y,
+    sample_domain=None,
+    allow_source: bool = True,
+    allow_multi_source: bool = True,
+    allow_target: bool = True,
+    allow_multi_target: bool = True,
+    # xxx(okachaiev): most likely this needs to be removed as it doesn't fit new API
```
kachayev, line 27: Better remove this comment as it definitely outdated :)

skada/utils.py
```diff
@@ -0,0 +1,247 @@
...
+def check_X_domain(
+    X,
+    sample_domain,
+    *,
+    allow_domains: Optional[Set[int]] = None,
+    allow_source: bool = True,
+    allow_multi_source: bool = True,
+    allow_target: bool = True,
+    allow_multi_target: bool = True,
+    # xxx(okachaiev): most likely this needs to be removed as it doesn't fit new API
```
kachayev, line 119: Same here, please, remove the comment

skada/utils.py
```diff
@@ -0,0 +1,247 @@
...
+        Allow the presence of target domains.
+    allow_multi_target : bool, optional (default=True)
+        Allow multiple target domains.
+    allow_auto_sample_domain : bool, optional (default=True)
+        Allow automatic generation of sample_domain if not provided.
+
+    Returns:
+    ----------
+    X : array
+        Combined input features for source and target domains.
```
kachayev, line 149: Should it be just 'input features' the same way it is marked in the parameters section?

skada/utils.py
```diff
@@ -0,0 +1,247 @@
...
+        Combined domain labels for source and target domains.
+    """
+    X = check_array(X, input_name='X')
+
+    if sample_domain is None and not allow_auto_sample_domain:
+        raise ValueError("Either 'sample_domain' or 'allow_auto_sample_domain' "
+                         "should be set")
+    elif sample_domain is None and allow_auto_sample_domain:
+        # default target domain when sample_domain is not given
+        # xxx(okachaiev): I guess this should be -inf instead of a number
```
kachayev, line 160: This comment is also outdated

kachayev (changes requested):

skada/tests/test_utils.py
YanisLalou: Yes it does, because you prefer to have the parentheses being right below the first letter of 'match' ?

kachayev: I would expect closing parentheses to be aligned with the beginning of 'match', yes.

kachayev (approved):

PR #54, opened 2024-01-12
Flake8 correction for _samples_generator.py

YanisLalou:


kachayev:
I thought we have a check in place that prevents us from merging before Flake8 is happy 🤔

Issue #67, opened 2024-01-31
[BUG] ValueError raised when using the GaussianReweightDensityAdapter


YanisLalou:
Using this code snippet:
```
from skada import GaussianReweightDensityAdapter, make_da_pipeline
from sklearn.decomposition import PCA
from sklearn.linear_model import LogisticRegression
from skada.datasets import fetch_office31_surf_all

domain_dataset_office31 = fetch_office31_surf_all()

pipeline = make_da_pipeline(
    GaussianReweightDensityAdapter(),
)

X, y, sample_domain = domain_dataset_office31.pack_lodo()
pipeline.fit(X=X, y=y, sample_domain=sample_domain)
```

You'll get this error:
`ValueError: Found array with 0 sample(s) (shape=(0, 800)) while a minimum of 1 is required by StandardScaler.
`

Also its worth noting that by adding a `LogisticRegression` at the end of the pipeline, the error magically disappears.

tgnassou:
I think the `pipeline` treats differently the last step of a pipeline and the steps before. That's why adding `LogisticRegression` fixes the problem, I think. When the adapter is the last step, the `pipeline` function seems to remove the negative values (i.e., the target) in `sample_domain`, but I don't know why.

kachayev:
The transform removes target values because it 'prepares' the output for the next step, which is supposed to be an estimator that doesn't have the access to targets (as we can't fit to masked labels).

kachayev:
The selector, not the transformer, sorry for the confusion. We can prevent this from happening by trying to guess what type of estimator is used. But I'm not sure how deep we want to go with that.

YanisLalou:
We could just check if the estimator has a 'transform' method --> we don't remove the negative values.
If the estimator has a 'predict' method --> we remove the negative values.

kachayev:
Yes, this is how default sklearn pipeline makes the distinction.

PR #68, opened 2024-01-31
[TO_REVIEW] Using global variables instead of number

YanisLalou:
+ Using global variables instead of number
+ Solution for Issue #67 (Add a condition to mask labels depending on the existence of a transform method or not)


skada/tests/test_base.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    pipe = make_da_pipeline(
+        LogisticRegression(),
+    )
+
+    selector = pipe['logisticregression']
+    print(selector.get_params())
+    X_output, y_output, routed_params = selector._remove_masked(
+        X, y, selector.get_params()
```
kachayev, line 31: Why do we need to pass parameters if `_removed_masked` is called on the instance of the selector?

skada/_utils.py
```diff
@@ -68,30 +68,61 @@ def _check_y_masking(y):
...
+    elif y_type == 'classification':
+        if (np.any(y < _DEFAULT_MASKED_TARGET_CLASSIFICATION_LABEL) or
+                not np.any(y == _DEFAULT_MASKED_TARGET_CLASSIFICATION_LABEL)):
+            raise ValueError("For a classification task, "
+                             "masked labels should be, "
+                             f"{_DEFAULT_MASKED_TARGET_CLASSIFICATION_LABEL}")
+        else:
+            return y_type
+    else:
+        raise ValueError("Uncompatible label type: %r" % y_type)
```
kachayev, line 90: Let's do string formatting with `f""` everywhere to stay consistent.

skada/tests/test_base.py
YanisLalou: In the `.fit()` function of the selector, we modify `routed_params` depending on the input `X`. Thus we need to feed it to `_remove_masked()` since `routed_params` could potentially be different than `self.get_params()` in that case.

Thus here I just called the function `_remove_masked()` as it is called in the `fit()` function.

kachayev:
This one is ready for merge as well, right? @YanisLalou

PR #70, opened 2024-02-02
[TO_REVIEW] Add test cases for the Reweight class

YanisLalou:
Super quick to review
Add test cases to make codevoc happy

### DIFF
--- a/skada/tests/test_reweight.py
+++ b/skada/tests/test_reweight.py
@@ -52,3 +52,27 @@ def test_reweight_estimator(estimator, da_dataset):
     assert np.mean(y_pred == y_test) > 0.9
     score = estimator.score(X_test, y_test, sample_domain=sample_domain)
     assert score > 0.9
+
+
+def test_reweight_gamma_float(da_dataset):
+    X_train, y_train, sample_domain = da_dataset.pack_train(
+        as_sources=['s'],
+        as_targets=['t']
+    )
+
+    estimator = KLIEPAdapter(gamma=0.2)
+    estimator.fit(X_train, y_train, sample_domain=sample_domain)
+
+
+def test_reweight_warning(da_dataset):
+    X_train, y_train, sample_domain = da_dataset.pack_train(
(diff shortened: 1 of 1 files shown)


skada/tests/test_reweight.py
```diff
@@ -52,3 +52,27 @@ def test_reweight_estimator(estimator, da_dataset):
...
     assert score > 0.9
+
+
+def test_reweight_gamma_float(da_dataset):
+    X_train, y_train, sample_domain = da_dataset.pack_train(
+        as_sources=['s'],
+        as_targets=['t']
+    )
+
+    estimator = KLIEPAdapter(gamma=0.2)
```
kachayev, line 63: I thought we had a large test case for all adapters, maybe just add this as a additional case there?

skada/tests/test_reweight.py
```diff
@@ -52,3 +52,27 @@ def test_reweight_estimator(estimator, da_dataset):
...
+def test_reweight_warning(da_dataset):
+    X_train, y_train, sample_domain = da_dataset.pack_train(
+        as_sources=['s'],
+        as_targets=['t']
+    )
+
+    estimator = KLIEPAdapter(gamma=0.1, max_iter=0)
+    estimator.fit(X_train, y_train, sample_domain=sample_domain)
+
+    with pytest.warns(UserWarning,
```
kachayev, line 76: Why do we want to check that it doesn't converge?

YanisLalou: Just wanted to test that the UserWarning was raised if n_iter > max_iter.
That's why in the test max_iter=0, so that we're sure we're in that case

skada/tests/test_reweight.py
YanisLalou: Yes sure

skada/tests/test_reweight.py
kachayev: Ah, I see

Issue #61, opened 2024-01-26
Can't use SupervisedScorer when using X, y, domain = dataset.pack_lodo


YanisLalou:
When using `LeaveOneDomainOut` we want to have a cross validation by modifying the source and target domains at each split.
Thus we need to use the `dataset.pack_lodo` method since its the only one not asking for a source and a target domain.

Now if we want to use the `SupervisedScorer`, it asks a `target_labels` argument to generate a score.
However when we use the `pack_lodo` function we don't have the `target_labels`.

Thus we can't use `SupervisedScorer` when using `LeaveOneDomainOut`. 
And by leaving the param `target_labels` empty in cross_validate , we get ValueErrors.

kachayev:
Getting `target_labels` for `SupervisedScorer` when using `pack_lodo` is possible, it's just a little bit tedious as it requires understanding of the packing. You need to take unmasked labels and duplicate it into masked positions. Something like

```
y_source, y_target = source_target_split(y, sample_domain=sample_domain)
target_labels = source_target_merge(y_source, y_source, sample_domain=sample_domain)
```

`target_labels` should now be passed to the fit params. The rest will happen automatically (the correct labels will be propagated through the entire machinery of the model selection). For now, I would rather put this in the document/examples for the `SupervisedScorer`, rather than keeping any sort of functionality dedicated to this (edge-)case.

YanisLalou:
Right now it doesn't work since `_merge_source_target()` expects a 2D array as an input to work.
Let me change that, so that we can feed `y` to it.

kachayev:
I think it would be also better to maintain parity between 'split' and 'merge', so it would look like 'source_target_merge' rather than an old name. Also, should be public in utils. We can do it here or in a separate PR, as you can see fit better.

PR #71, opened 2024-02-02
[TO_REVIEW] Modification source_target_merge function behaviours

YanisLalou:
Issue: #61 
Change the name of `source_target_merge`
Allow an N dimension array as an input for `source_target_merge`
Add test cases

For `source_target_merge`:
- Sample domain input should be optional
- Y_target can be equal to None (check there's only one None given in all pairs)
- X_source/target can be list of arrays, check all sizes are coherent


skada/utils.py
```diff
@@ -243,3 +243,56 @@ def source_target_split(
...
+    if X_source.shape[0] > 0 and X_target.shape[0] > 0:
+        output = np.zeros_like(
+            np.concatenate((X_source, X_target)), dtype=X_source.dtype
+        )
+        output[sample_domain >= 0] = X_source
+        output[sample_domain < 0] = X_target
+
+    elif X_source.shape[0] > 0:
+        output = np.zeros_like(
+            X_source, dtype=X_target.dtype
```
tgnassou, line 288: `dtype=X_source.dtype`

tgnassou: For `X_target` a empty set I don't know what will be the `dtype`

tgnassou (approved):

skada/utils.py
```diff
@@ -243,3 +243,56 @@ def source_target_split(
     return list(chain.from_iterable(
         (a[source_idx], a[~source_idx]) for a in arrays
     ))
+
+
+def source_target_merge(
```
kachayev, line 248: I think we should match split API: accept as many pairs of 'things' as necessary, alongside with a named argument 'sample_domain'. 

So I can do something like

```python 
X, y = source_target_merge(X_source, X_target, y_source, y_target, sample_domain=sample_domain)
```

kachayev (changes requested):

skada/utils.py
rflamary: Do you mean : 

```python
X, y, sample_domain = source_target_merge(X_source, X_target, y_source, y_target)
```

insterad?

YanisLalou: In `source_target_split()` we don't output back the sample domain, just the splitted arrays. Makes sense to do the same here

kachayev: Right, as we have this function to pair up with the split functionality, we shouldn't encourage swapping sample_domain to a new one, even though it's technically possible (it would mean that we have to return it with every operation). Moreover, it might easily lead to 'accidental' loss of the information when having multiple sources/targets.

YanisLalou:
For now `source_target_merge()` accepts an even number of arrays as input + the sample domain array.
It works by considering the arrays by duo (the first one is considered as the source and the second the target).
Thus I'm wondering:
+ if it would be better to accept the duos as tuples rather than directly in the *arrays.
+ Don't know if the doc is clear enough 

Would love your feedbacks

kachayev:
Regarding pairs, you are right, it might be better with pairs. But it's common to have pairs of ndarrays as an input type... So it might also be a bit confusing. Let's keep the current implementation to see how it works in practice. We will be able to change it if necessary.

kachayev:
@YanisLalou is this one ready for review?

YanisLalou:
> @YanisLalou is this one ready for review?

Almost, there's one codevoc test that I still need to implement + check if I've managed to handle all possible cases.
But otherwise yes you can start reviewing it + if you find any edge cases that I didn't foreseen, that would really help me

skada/tests/test_utils.py
```diff
@@ -290,3 +294,126 @@ def test_extract_source_indices():
...
+        pos_source=0.1,
+        pos_target=0.9,
+        n_samples_source=n_samples_source,
+        n_samples_target=n_samples_target,
+        random_state=0,
+        return_X_y=True,
+    )
+
+    X_source, X_target = source_target_split(X, sample_domain=sample_domain)
+    y_source, y_target = source_target_split(y, sample_domain=sample_domain)
```
kachayev, line 312: Shouldn't `source_target_split` function accept both `X, y` together?

skada/tests/test_utils.py
```diff
@@ -290,3 +294,126 @@ def test_extract_source_indices():
...
+    # Test y_target = None + Consistent number of samples in source-target
+    _ = source_target_merge(
+        X_source,
+        X_target,
+        y_source,
+        None,
+        sample_domain=sample_domain
+    )
+
+    # Test 2 None in a duo arrays
```
kachayev, line 402: 'array pairs' or 'pairs of arrays'

skada/utils.py
```diff
@@ -247,3 +251,223 @@ def source_target_split(
...
+    sample_domain : Optional[np.ndarray] = None
+) -> np.ndarray:
+    f""" Merge source and target domain data based on sample domain labels.
+
+    Parameters
+    ----------
+    *arrays : sequence of array-like
+        (, n_features). The number of arrays must be even
+        since we consider duos of source-target arrays each time.
+        Each duo should have at least one non-empty array.
```
kachayev, line 267: Each pair

skada/utils.py
```diff
@@ -247,3 +251,223 @@ def source_target_split(
...
+    # ################ Check sample_domain #################
+    # If sample_domain is None, we need to infer it from the target array
+    if sample_domain is None or sample_domain.shape[0] == 0:
+        # We need to infer the domain from the target array
+        warnings.warn(
+            "sample_domain is None or empty, it will be inferred from the arrays"
+        )
+
+        # By assuming that the first array is the source and the second the target
+        index_something_source = 0
```
kachayev, line 357: Why `something`?

skada/utils.py
```diff
@@ -247,3 +251,223 @@ def source_target_split(
...
+            y_type = type_of_target(arrays[duo_index])
+            if y_type == 'binary' or y_type == 'multiclass':
+                arrays[index_is_empty] = (
+                    _DEFAULT_MASKED_TARGET_CLASSIFICATION_LABEL *
+                    np.ones(
+                        (sample_domain.shape[0] - arrays[duo_index].shape[0],) +
+                        arrays[duo_index].shape[1:]
+                    )
+                )
+            else:
```
kachayev, line 404: The only difference between it and else branches is the value of the label. Thus the code could be simplified.

kachayev (changes requested):
Overall comment, let's replace 'duo' with 'pair'. I mentioned it multiple times in different places in the code but certainly not everywhere.

skada/utils.py
```diff
@@ -247,3 +253,219 @@ def source_target_split(
         (a[source_idx], a[~source_idx]) if a is not None else (None, None)
         for a in arrays
     ))
+
+
+def source_target_merge(
+    *arrays,
+    sample_domain : Optional[np.ndarray] = None
+) -> np.ndarray:
```
kachayev, line 261: Missed this one last time: the return type should be a tuple of ndarrays, moreover, of unknown size. `Sequence` would be probably the closest container type for such situations.

kachayev (changes requested):

PR #76, opened 2024-02-07
[MRG] JDOT Regressor

rflamary:
New JDOTRegressor class


skada/_ot.py
```diff
@@ -0,0 +1,185 @@
+# Author: Remi Flamary <remi.flamary@polytechnique.edu>
```
kachayev, line 1: Should we move OT mapping-based solvers into _ot module as well? (not in this PR). It might be better to keep all of them together to make `POT` into an optional dependency.

skada/_ot.py
```diff
@@ -0,0 +1,185 @@
...
+
+        res = solve_jdot_regression(Xs, ys, Xt, self.base_estimator,
+                                    alpha=self.alpha, n_iter_max=self.n_iter_max,
+                                    tol=self.tol, verbose=self.verbose, **self.kwargs)
+
+        self.estimator_, self.lst_loss_ot_, self.lst_loss_tgt_labels_, self.sol_ = res
+
+    def predict(self, X, sample_domain=None, *, sample_weight=None):
+        """Predict using the model"""
+        return self.estimator_.predict(X)
```
kachayev, line 185: We need `check_if_fitted` here

skada/_ot.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(self, base_estimator, alpha=0.5, n_iter_max=100,
+                 tol=1e-5, verbose=False, **kwargs):
+        self.base_estimator = base_estimator
+        self.kwargs = kwargs
+        self.alpha = alpha
+        self.n_iter_max = n_iter_max
+        self.tol = tol
+        self.verbose = verbose
+
+    def fit(self, X, y=None, sample_domain=None, *, sample_weight=None):
```
kachayev, line 172: With the routing mechanism, I guess there's no need to specify `sample_weight` in the list of arguments if don't use those.

skada/base.py
```diff
@@ -119,6 +119,25 @@ def transform(
         )
 
 
+class DAEstimator(BaseEstimator):
+    """ Generic DA estimator class
+
+    """
+
+    __metadata_request__fit = {'sample_domain': True}
+    __metadata_request__transform = {'sample_domain': True, 'allow_source': True}
```
kachayev, line 128: This should be `predict`, right? The estimator doesn't provide `transform` without a special mixin

kachayev, line 128: I guess we also need to specify request for `score`, `decision_function`, `predict_proba`

rflamary: Very good point  and one of the reasons why I want a class so that contributor do not need to worry about that ;)

skada/_ot.py
rflamary: I do not use it yet but I wanted to implement with weights in next commits, Thanks for reminding me.

skada/_ot.py
rflamary: that's an idea but indeed out of teh csope of this PR

skada/_ot.py
rflamary: Done I needed to update `split_domain` function to handle None samples_weight

skada/_ot.py
```diff
@@ -0,0 +1,213 @@
...
+    ws : array-like of shape (n_samples,)
+        Source domain weights (will ne normalized to sum to 1).
+    wt : array-like of shape (m_samples,)
+        Target domain weights (will ne normalized to sum to 1).
+    n_iter_max: int
+        Max number of JDOT alternat optimization iterations.
+    tol: float>0
+        Tolerance for loss variations (OT and mse) stopping iterations.
+    verbose: bool
+        Print loss along iterations if True.as_integer_ratio
```
kachayev, line 45: 'kwargs' is not documented (it is common to pass additional arguments into a base estimator but still better to mention explicitly)

skada/_ot.py
```diff
@@ -0,0 +1,213 @@
...
+
+    References
+    ----------
+    [1] N. Courty, R. Flamary, A. Habrard, A. Rakotomamonjy, Joint Distribution
+        Optimal Transportation for Domain Adaptation, Neural Information Processing
+        Systems (NIPS), 2017.
+
+    """
+
+    estimator = clone(base_estimator)
```
kachayev, line 67: If we return 'estimator' first, should we also take it as the first argument?

skada/_ot.py
```diff
@@ -0,0 +1,213 @@
...
+        res = solve_jdot_regression(Xs, ys, Xt, self.base_estimator,
+                                    alpha=self.alpha, n_iter_max=self.n_iter_max,
+                                    tol=self.tol, verbose=self.verbose, **self.kwargs)
+
+        self.estimator_, self.lst_loss_ot_, self.lst_loss_tgt_labels_, self.sol_ = res
+
+    def predict(self, X, sample_domain=None, *, sample_weight=None):
+        """Predict using the model"""
+        check_is_fitted(self)
+        if sample_domain is not None and np.any(sample_domain < 0):
```
kachayev, line 200: Source domains are > 0, right? 🤔

PR #79, opened 2024-02-07
[to_review] parameter for 2d data generation with 'regression' label

BuenoRuben:
it allows to choose mean and covariance matrix for the generated data


rflamary:
While you are at it @BuenoRuben could you please scale the output y so that it has reasonable values? it is very small at the moment so that when i computed MSE I had something around 1e-6

It should have a variance around 1 if possible (but do not scale it for each seed it would introduce some other shift)

BuenoRuben:
i am doing it right now

BuenoRuben:
> While you are at it @BuenoRuben could you please scale the output y so that it has reasonable values? it is very small at the moment so that when i computed MSE I had something around 1e-6
> 
> It should have a variance around 1 if possible (but do not scale it for each seed it would introduce some other shift)

it should be fine now

tgnassou:
I don't know what happened, but your examples show that label distributions are changing between source and target. Look at Rémi's DA course. There is an example of a regression dataset for covariate shift etc. Maybe it can help you to see what it should look like.

BuenoRuben:
I noticed, it is due to the new scale, but when I tried to scale them accordingly in the example it was not working either, it is normally just because of the choice of the cmap, I will change that now it pause a problem

BuenoRuben:
it should be fixed now, we have only two colors for the y-values (for high and low values), this thus should solved the problem

BuenoRuben:
that's corrected

skada/datasets/_samples_generator.py
```diff
@@ -25,7 +25,10 @@ def _generate_unif_circle(n_samples, rng):
     return x
 
 
-def _generate_data_2d_classif(n_samples, rng, label='binary'):
+def _generate_data_2d_classif(
+        n_samples, rng, mu_regression=np.array([0, 0]),
+        sigma_regression=np.array([[1, 0], [0, 1]]),
```
rflamary, line 30: never put mutable variable (such as lists or numpy arrays) as default parameter of a function. Use Noe and set the value to default is None in the function.

Issue #43, opened 2024-01-09
Implement DA methods from "DA under Target and Conditional Shift"


kachayev:
Link to the paper: https://proceedings.mlr.press/v28/zhang13d.pdf

PR #93, opened 2024-02-16
[MRG] Add TarS method

antoinecollas:
This PR implements the TarS method from  http://proceedings.mlr.press/v28/zhang13d.pdf (mainly eq.6).
Related issue: #43


skada/_reweight.py
```diff
@@ -7,6 +7,7 @@
 import warnings
 
 import numpy as np
+from cvxopt import matrix, solvers
```
YanisLalou, line 10: why not using SciPy minimizer ?

skada/_reweight.py
```diff
@@ -669,3 +673,192 @@ def KLIEP(
...
+                err_msg="Source data is different from the one used"
+                " for fitting and new labels are not provided."
+            )
+            y_source = self.y_source_fit_
+        else:
+            X, y = check_X_y_domain(X, y, sample_domain)
+            y_source, _ = source_target_split(y, sample_domain=sample_domain)
+
+        # check y is discrete or continuous
+        discrete = np.issubdtype(y_source.dtype, np.integer)
```
YanisLalou, line 782: maybe better to use either the` _find_y_type()` from `_utils.py `or `type_of_target()` from sklearn because some labels can be float but still discrete

skada/_reweight.py
antoinecollas: The problem TarS solves is a quadratic problem under constraints. I don't think SciPy can solve this out of the box and CVXOPT is made for this .

YanisLalou: Look at #94, the KMM method solves also a quadratic problem under constraints but with SciPy. But tbh I havent yet looked at the TarS paper thoroughly, so maybe I'm missing something. Also if you think that CVXOPT is better for this kind of problem, yeah sure go ahead

tgnassou: If it is possible it could be nice to be consistent in the toolbox and not to add a new requirement

skada/_reweight.py
```diff
@@ -669,3 +673,192 @@ def KLIEP(
...
+        m = X_source.shape[0]
+        n = X_target.shape[0]
+
+        # if no source data, return uniform weights for target
+        if m == 0:
+            return AdaptationOutput(X=X, y=y, sample_weights=np.zeros(n))
+
+        # if no y provided, get the source labels from fit method
+        if y is None:
+            np.testing.assert_array_equal(
```
tgnassou, line 770: Following the PR #94, the optimization is done in the `.fit` and update if `X_source != self.X_source_fit_` is it possible to have the same ?

tgnassou (commented):
I checked briefly the optimisation problem, and everything was good

antoinedemathelin:
Hello,
Yes, solving constrained QP is possible with scipy [link](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html). There are three available solvers for that: `trust-constr`, `SLSQP` and `COLYBA`. I tested them, and `SLSQP` is the fastest. Giving the Jacobian (computed analytically) fasten the computation also.

Indeed, CVXOPT is faster (perhaps 10 times faster than scipy for 1000 samples). However, I do not recommend including it in the requirements. I did it for adapt, but run in multiple bugs after new cvxopt releases...

My suggestion is to provide a faster optimization solver, using the Frank-Wolfe algorithm, which can be easily implemented. Frank-Wolfe works well for weighting methods based on MMD or KL minimization, such as KLIEP and KMM. It really speeds up the optimization. cf this paper : https://webdocs.cs.ualberta.ca/~dale/papers/ijcai15.pdf . It could be proposed as an alternative for the user via a `solver` argument, to keep track of the original algorithm that uses classic QP solvers.

antoinecollas:
Thanks for all the infos @antoinedemathelin !

skada/_reweight.py
```diff
@@ -669,3 +674,223 @@ def KLIEP(
...
+
+        # compute M
+        K_cross = pairwise_kernels(X_target, X_source, metric="rbf", gamma=self.gamma)
+        M = np.ones((1, n)) @ K_cross @ omega.T
+
+        # solve the optimization problem
+        # min_alpha 0.5 * alpha^T P alpha - q^T alpha
+        # s.t. 0 <= R alpha <= B_beta
+        #      m (1 - eps) <= 1^T R alpha <= m (1 + eps)
+        P = 0.5 * (R.T @ A @ R)
```
YanisLalou, line 750: 1/2 here is redundant with the 1/2 in func(alpha), line 751

antoinecollas: Good catch! Thank you

skada/_reweight.py
```diff
@@ -669,3 +674,223 @@ def KLIEP(
...
+        )
+
+        source_idx = extract_source_indices(sample_domain)
+
+        if source_idx.sum() > 0:
+            if np.array_equal(self.X_source_, X[source_idx]):
+                source_weights = self.weights_
+            else:
+                # get the nearest neighbor in the source domain
+                K = pairwise_kernels(
```
rflamary, line 854: why use a kernel with badwidth here? if you use the nearest neighbor then you should just use distance?

skada/_reweight.py
```diff
@@ -669,3 +674,223 @@ def KLIEP(
...
+
+        # compute M
+        K_cross = pairwise_kernels(X_target, X_source, metric="rbf", gamma=self.gamma)
+        M = np.ones((1, n)) @ K_cross @ omega.T
+
+        # solve the optimization problem
+        # min_alpha 0.5 * alpha^T P alpha - q^T alpha
+        # s.t. 0 <= R alpha <= B_beta
+        #      m (1 - eps) <= 1^T R alpha <= m (1 + eps)
+        P = R.T @ A @ R
```
rflamary, line 750: sems like a generic QP solver will be usefull elsewhere. Do we want to code a function that takes (we can put that in utils) standard QP as input?

skada/_reweight.py
```diff
@@ -669,3 +674,223 @@ def KLIEP(
...
+
+
+class MMDTarSReweightAdapter(BaseAdapter):
+    """MMDTarSReweight adapter.
+
+    The idea of MMDTarSReweight is to find an importance estimate \beta(y) such that
+    the Maximum Mean Discrepancy (MMD) divergence between the source input density
+    p_source(x) to its estimate p_target(x) is minimized under the assumption
+    of equal conditional distributions p(x|y) for both source and target domains.
+
```
rflamary, line 686: Add discussion here (warning block) about how the adapter works and previsdes reweighting for source samples with nearest neighbor else does nothing

skada/_reweight.py
```diff
@@ -669,3 +674,223 @@ def KLIEP(
...
+                source_weights = self.weights_[idx]
+            source_idx = np.where(source_idx)
+            weights = np.zeros(X.shape[0], dtype=source_weights.dtype)
+            weights[source_idx] = source_weights
+        else:
+            weights = None
+        return AdaptationOutput(X=X, sample_weight=weights)
+
+
+def MMDTarSReweight(base_estimator=None, gamma=1.0):
```
rflamary, line 870: add all parameter from the adapter + doc

skada/tests/test_reweight.py
```diff
@@ -61,6 +69,27 @@ def test_reweight_estimator(estimator, da_dataset):
     assert score > 0.9
 
 
+@pytest.mark.parametrize(
```
rflamary, line 72: either this test can be done with paramatrizing other estimator (or adding the estimator in another test) or it should have a specifics purpose and less generic name.

rflamary (commented):
looks good but a few comments below.

skada/tests/test_reweight.py
antoinecollas: It tests the TarS method on a regression problem. Sorry, I don't get the problem...

skada/utils.py
```diff
@@ -515,8 +515,9 @@ def _merge_arrays(
     return output
 
 
-def qp_solve(Q, c=None, A=None, b=None, Aeq=None, beq=None, lb=None, ub=None,
-             x0=None, tol=1e-6, max_iter=1000, verbose=False, log=False):
+def qp_solve(Q, c=None, A=None, Alb=None, Aub=None, lb=None, ub=None,
+             Aeq=None, beq=None, x0=None, tol=1e-6, max_iter=1000,
+             verbose=False, log=False):
```
antoinedemathelin, line 520: Sorry @antoinecollas, but I prefer the previous version of `qp_solve`, using a constraint Ax >= b can be done by using opposite sign factors in A and b. What do you think ?

antoinedemathelin: Besides, `lb` and `ub` should be used for constraints `lb <= x <= ub`, I think

antoinedemathelin: But maybe you are right, I am not really sure what is the best implementation here... @rflamary what do you think?

rflamary: it is more standard to have a unique A,b for inequality Ax <=b (lower not upper). I agree that we should do that (sorry @antoinecollas )

rflamary: I lke Aeq, beq and A,b for equality an inequality also if we decide to cange it I think it is easier to read from teh code

rflamary (approved):

rflamary (approved):

Issue #82, opened 2024-02-09
Implement Kernel Mean Matching


antoinedemathelin:
Hi everyone,
I am glad to see that the skada repo is already public. I really like the API choices that have been made (the pipeline idea is great!)
I am opening this issue to propose the implementation of the Kernel Mean Matching reweighting method (cf. “Correcting sample selection bias by unlabeled data.” [paper](https://proceedings.neurips.cc/paper_files/paper/2006/file/a2186aa7c086b46ad4e8bf81e2a3a19b-Paper.pdf).

Are you ok to add it to the library ? If yes, I can open a PR.

rflamary:
Hello @antoinedemathelin and welcome back to skada !. You have already coded many methods that are of interest to us.

This is a very good idea, KMM is a reweighting isn't it? you could do it with a specific Adapter that outputs sample_weights ;)

antoinedemathelin:
Thank you @rflamary !
Yes it's a reweighting algorithm. From what I understand, I should write a `KMMAdapter` object, implementing an `adapt` method which returns an `AdaptationOutput` instance containing the sample_weights of the reweighting. If I am right, the inputs arrays `X, y`, now contain both source and target samples, so the `sample_weights` array should contain the source importance weights for the source indexes and 0 for the target indexes ? I will follow the `KLIEP` skada implementation, the main difference with KMM is the optimization part, the "API part" should be the same, I guess.

I guess, I need to write a small example and tests for the method ?

tgnassou:
Exactly! If you have any problem, just ask us. In the example don't hesitate to show how works the method with nice plots ;) 

And don't hesitate to give us some feedback on the API, we are still converging to the final API

PR #94, opened 2024-02-16
[MRG] Add KMM method + example

antoinedemathelin:
Hi everyone,
I open this PR to propose an implementation of the KMM method from the paper https://proceedings.neurips.cc/paper_files/paper/2006/file/a2186aa7c086b46ad4e8bf81e2a3a19b-Paper.pdf as suggested in issue #82.

KMM has the particularity to directly learn the source weights corresponding to the source data Xs given in `fit`. In the `adapt` method, I follow the same idea as for OT. The source data matrix are stored in an attribute `X_source_`, then, in the adapt method, we compare the given X[src_idx] with X_source_ (cf. [OT1](https://github.com/PythonOT/POT/blob/c84ef332ea50a5f94a72f435a7828989c15fa4b5/ot/da.py#L593)), if both are equal, we return the weights learned in fit. If not, we return the weight of the closest neighbor in X_source_, as done for OT mapping, cf. [OT2](https://github.com/PythonOT/POT/blob/c84ef332ea50a5f94a72f435a7828989c15fa4b5/ot/da.py#L612) (note that we use the kernel function to find the closest neighbor).


examples/methods/plot_kmm_reweighting.py
```diff
@@ -0,0 +1,60 @@
...
+====================
+This example illustrates the use of KMM method [1] to correct covariate-shift.
+.. [1] J. Huang, A. Gretton, K. Borgwardt, B. Schölkopf and A. J. Smola.
+       'Correcting sample selection bias by unlabeled data.'
+       In NIPS, 2007.
+"""
+
+# Author: Antoine de Mathelin
+#
+# License: MIT License
```
YanisLalou, line 12: We've changed the license a while back I think, now its the BSD 3-Clause

examples/methods/plot_kmm_reweighting.py
```diff
@@ -0,0 +1,60 @@
...
+# %%
+# Generate sample bias dataset
+# ----------------------------
+np.random.seed(123)
+Xs = np.random.randn(100, 2) * .5
+Xt = np.random.random((20, 2))
+
+X = np.concatenate((Xs, Xt))
+sample_domain = np.ones(X.shape[0])
+sample_domain[Xs.shape[0]:] *= -2
```
YanisLalou, line 29: Usually we use the `make_shifted_datasets()` method  to generate our example datasets. Here you can use it with the arg `shift="covariate_shift"`

examples/methods/plot_kmm_reweighting.py
```diff
@@ -0,0 +1,65 @@
...
+        noise=0.4,
+        random_state=123,
+        return_dataset=True
+)
+X, y, sample_domain = ds.pack_train(as_sources=['s'], as_targets=['t'])
+Xs, ys = ds.get_domain("s")
+Xt, yt = ds.get_domain("t")
+
+# %%
+# Illustration of Importance Weighting
```
rflamary, line 37: Very nice, maybe we could also illustrate the trained predictor with and woithout teh revwieghting? after tgis visualisation?

skada/_reweight.py
```diff
@@ -8,8 +8,9 @@
 
 import numpy as np
 from scipy.stats import multivariate_normal
+from scipy.optimize import minimize, LinearConstraint
```
rflamary, line 11: We will have  a conflict between antoines here (@antoinedemathelin VS @antoinecollas ) ;)

skada/_reweight.py
```diff
@@ -669,3 +670,272 @@ def KLIEP(
...
+    B : float, default=1000.
+        Weight upper bound.
+    eps : float, default=None
+        KMM tolerance parameter. If `None`, eps is set to
+        (sqrt(n_samples_source) - 1) / sqrt(n_samples_source).
+    tol : float, default=1e-6
+        Tolerance for the stopping criterion in the optimization.
+    max_iter : int, default=100
+        Number of maximum iteration before stopping the optimization.
+    random_state : int, RandomState instance or None, default=None
```
rflamary, line 704: why random state hare?

skada/_reweight.py
```diff
@@ -669,3 +670,272 @@ def KLIEP(
...
+        """
+        X, sample_domain = check_X_domain(
+            X,
+            sample_domain,
+            allow_multi_source=True,
+            allow_multi_target=True
+        )
+        X_source, X_target = source_target_split(X, sample_domain=sample_domain)
+
+        self.weights_ = self._weights_optimization(X_source, X_target)
```
rflamary, line 774:
```suggestion
        self.source_weights_ = self._weights_optimization(X_source, X_target)
```
maybe?

skada/_reweight.py
```diff
@@ -669,3 +670,272 @@ def KLIEP(
...
+                               degree=self.degree,
+                               coef0=self.coef0)
+        Kst = pairwise_kernels(X_source,
+                               X_target,
+                               metric=self.kernel,
+                               filter_params=True,
+                               gamma=self.gamma,
+                               degree=self.degree,
+                               coef0=self.coef0)
+        Ns = Kss.shape[0]
```
rflamary, line 794: Same comment as is did to @antoinecollas , we need a function for QP solver, we will use it many times probably. You should proably work together on this one with one merge before the other

skada/_reweight.py
```diff
@@ -669,3 +670,272 @@ def KLIEP(
...
+        )
+
+        source_idx = extract_source_indices(sample_domain)
+
+        if source_idx.sum() > 0:
+            if np.array_equal(self.X_source_, X[source_idx]):
+                source_weights = self.weights_
+            else:
+                # get the nearest neighbor in the source domain
+                K = pairwise_kernels(X[source_idx], self.X_source_,
```
rflamary, line 860: Same cmment again, why compute a kerel when distane is enough?

skada/_reweight.py
rflamary: Also I'm sorry but you estimate a kernel machine reweighting sum_i w_i k(x,,x_i) my intuition is you should compute this rewieghting instead of juste the weights estimated on the points and then nearest neighbor (let us talk about this in slack)

rflamary (commented):
Left a few comments. We need to disus with @antoinecollas about a few things to avoid conflicts and repeated code

antoinedemathelin:
Hi everyone,
I added a `qp_solve` function to factorize the qp solver. I need to create specific test for it, I think.
@rflamary I added the `smooth_weights` parameter as we discussed.

rflamary (commented):
This is nice. We are missing a few tests on the new solver function and I would illustrate the final classifier and weights for smooth_weights and not smooth weights in the example (r elese nobody will use it)

examples/methods/plot_kmm_reweighting.py
```diff
@@ -0,0 +1,107 @@
+"""
+Kernel Mean Matching
+====================
+This example illustrates the use of KMM method [1] to correct covariate-shift.
```
rflamary, line 4: Small rendering problem in the example

image

skada/_reweight.py
```diff
@@ -669,3 +669,266 @@ def KLIEP(
...
+        self.kernel = kernel
+        self.gamma = gamma
+        self.degree = degree
+        self.coef0 = coef0
+        self.B = B
+        self.eps = eps
+        self.tol = tol
+        self.max_iter = max_iter
+        self.smooth_weights = smooth_weights
+        self.random_state = random_state
```
rflamary, line 746: random state is set but never used... remove it?

skada/_reweight.py
```diff
@@ -669,3 +669,266 @@ def KLIEP(
...
+                               gamma=self.gamma,
+                               degree=self.degree,
+                               coef0=self.coef0)
+        Ns = Kss.shape[0]
+        kappa = Ns * Kst.mean(axis=1)
+
+        if self.eps is None:
+            eps = (np.sqrt(Ns) - 1) / np.sqrt(Ns)
+        else:
+            eps = self.eps
```
rflamary, line 803: line ,never tested

skada/_reweight.py
```diff
@@ -669,3 +669,266 @@ def KLIEP(
...
+        )
+
+        source_idx = extract_source_indices(sample_domain)
+
+        if source_idx.sum() > 0:
+            if (np.array_equal(self.X_source_, X[source_idx])
+               and not self.smooth_weights):
+                source_weights = self.weights_
+            else:
+                K = pairwise_kernels(X[source_idx], self.X_source_,
```
rflamary, line 850: not covered by tests

skada/_reweight.py
```diff
@@ -669,3 +669,266 @@ def KLIEP(
...
+        self.coef0 = coef0
+        self.B = B
+        self.eps = eps
+        self.tol = tol
+        self.max_iter = max_iter
+        self.smooth_weights = smooth_weights
+        self.random_state = random_state
+
+        if kernel not in KERNEL_PARAMS:
+            raise ValueError("`kernel` argument should be included in %s,"
```
rflamary, line 749: Value Error not tested

skada/_reweight.py
rflamary: no problem if you want to keep weights but i think it is more clear

rflamary (commented):
I'm sorry @antoinedemathelin but O found a few stuff remaining see below. This is nearly perfect.

PR #108, opened 2024-02-26
 [MRG] Implementation of 1NN reweighting and reweighting example implementation

BuenoRuben:
based on this paper: https://arxiv.org/pdf/2102.02291.pdf


skada/_reweight.py
```diff
@@ -671,6 +673,145 @@ def KLIEP(
...
+        The estimator object fitted on the source data.
+    weight_estimator_target_ : object
+        The estimator object fitted on the target data.
+    """
+
+    def __init__(self, base_estimator=None, laplace_smoothing=False):
+        super().__init__()
+        self.laplace_smoothing = laplace_smoothing
+        self.base_estimator = base_estimator or KNeighborsClassifier(
+            n_neighbors=1)
```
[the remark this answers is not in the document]
antoinedemathelin: Something like :
`n_neighbors=1, weights='uniform', algorithm='auto', leaf_size=30, p=2, metric='minkowski', metric_params=None, n_jobs=None` that I would pass to a KNearestNeighborClassifier object

What do you think ?

skada/_reweight.py
```diff
@@ -671,6 +673,145 @@ def KLIEP(
...
+        return self
+
+    def get_weights(self, Xs, Xt):
+        indices_source = np.arange(Xs.shape[0])
+        estimator = clone(self.base_estimator)
+        estimator.fit(Xs, indices_source)
+        predictions = estimator.predict(Xt)
+        a = (1 if self.laplace_smoothing else 0)
+        weights = np.array(
+            [a + np.count_nonzero(predictions == i) for i in indices_source])
```
antoinedemathelin, line 739: Maybe you can fasten this part of the implementation by using `np.unique`, with something like:

```pyhton
unique, counts = np.unique(predictions, return_counts=True)
weights = np.ones(Xs.shape[0]) * float(self.laplace_smoothing)
weights[unique] += counts
```

skada/_reweight.py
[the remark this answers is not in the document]
BuenoRuben: yes you are right, I m going to use those parametters instead

skada/_reweight.py
antoinedemathelin: Hi @BuenoRuben,
I see that you mark this review as resolved, but it was not linked with my previous (not relevant) reviews for this part of the code. I still believe that using `np.unique` here is faster...

BuenoRuben:
@antoinedemathelin I added KMM to the reweighting methods' example, and noticed that when fitting with:
n_training, n_source = 20, 20; it's almost instantly fitted
n_training, n_source = 30, 30; it's taking some time
n_training, n_source = 50, 50; it's taking about 2 minutes, while others are taking less than 5 sec

I don't know if your method is specifically having hight complexity, so I just prefer to tell you

antoinedemathelin:
> @antoinedemathelin I added KMM to the reweighting methods' example, and noticed that when fitting with: n_training, n_source = 20, 20; it's almost instantly fitted n_training, n_source = 30, 30; it's taking some time n_training, n_source = 50, 50; it's taking about 2 minutes, while others are taking less than 5 sec
> 
> I don't know if your method is specifically having hight complexity, so I just prefer to tell you

Hi @BuenoRuben,
Yes, that's right, the qp solver used in KMM is pretty slow when n_source increase (note that n_source = 50 corresponds to 400 samples I think). This can be fasten by using `cvxopt`, which we will add as an option solver. I also plan to implement the Frank-Wolfe algorithm for KMM, which really speed up the optim.

As a first workaround, maybe we can reduce the default number of `max_iter` to 100 instead of 1000. The algorithm will not fully converge, but it will be faster...

skada/_reweight.py
```diff
@@ -675,6 +676,258 @@ def KLIEP(
...
+    pipeline : sklearn pipeline
+        Pipeline containing the ReweightDensity adapter and the base estimator.
+    """
+    if base_estimator is None:
+        base_estimator = LogisticRegression().set_fit_request(sample_weight=True)
+
+    return make_da_pipeline(
+        NearestNeighborDensityAdapter(
+            laplace_smoothing=laplace_smoothing,
+            weights="uniform",
```
tgnassou, line 919: You don't put the parameters of your function but hardcode the values

skada/_reweight.py
```diff
@@ -911,10 +1164,6 @@ def KMM(
         Number of maximum iteration before stopping the optimization.
     smooth_weights : bool, default=False
         If True, the weights are "smoothed" using the kernel function.
-
```
tgnassou, line 914: Why do you remove this? maybe a mistake

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,403 @@
...
+        alpha=1,
+        ax=ax,
+        eps=0.5,
+        response_method="predict",
+        plot_method="contour",
+    )
+
+    ax.set_xticks(()), ax.set_yticks(())
+    ax.set_xlim(x_min, x_max), ax.set_ylim(y_min, y_max)
+    ax.set_title("Training with rewegihted data", fontsize=12)
```
tgnassou, line 187: typo: rewegihted -> reweighted. And for withtout DA it is unweighted

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,403 @@
...
+#     Illustration of the Discr. reweighting method
+# ------------------------------------------
+#
+# This estimator derive a class of predictive densities by weighting the source samples
+# when trying to maximize the log-likelihood function. Such approach is effective in
+# cases of covariate shift.
+#
+# See [3] for details.
+#
+# .. [3]    Hidetoshi Shimodaira. Improving predictive inference under
```
tgnassou, line 258: Is it the same as [2]

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,403 @@
...
+weights = weight_estimator.adapt(X, sample_domain=sample_domain).sample_weight[idx]
+
+create_plots(clf, weights=weights, name="1NN Reweight Density")
+
+# %%
+#     Illustration of the Kernel Mean Matching method
+# ------------------------------------------
+# .. _Kernel Mean Matching
+#
+# This example illustrates the use of KMM method [5] to correct covariate-shift.
```
tgnassou, line 342: You cite [5] and [6] and maybe you could explain a bit more KMM

examples/methods/plot_reweighting.py
BuenoRuben: yes, in `_reweight.py` it is the same also

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,398 @@
...
+idx = extract_source_indices(sample_domain)
+weights = weight_estimator.adapt(X, sample_domain=sample_domain).sample_weight[idx]
+
+create_plots(clf, weights=weights, name="Reweight Density")
+
+# %%
+#     Illustration of the Gaussian reweighting method
+# ------------------------------------------
+# .. _Gaussian Reweight Density
+# See [1] for details.
```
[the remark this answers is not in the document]
rflamary: yes please

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,401 @@
...
+
+from skada import (
+    KLIEP,
+    KMM,
+    DiscriminatorReweightDensity,
+    GaussianReweightDensity,
+    ReweightDensity,
+    source_target_split,
+)
+from skada._reweight import NearestNeighborReweightDensity
```
rflamary, line 31: NearestNeighborReweightDensity should be imported in `skada.__init__.py`

examples/methods/plot_reweighting.py
```diff
@@ -0,0 +1,401 @@
...
+#
+# When not using domain adaptation, the classifier won't train on
+# data that is distributed as the target sample domain, it will thus
+# not be performing optimaly.
+
+# We create a dict to store scores:
+scores_dict = {}
+
+
+def create_plots(
```
rflamary, line 123:
```suggestion
def plot_weights_and_classifier(
``````

File: skada/__init__.py [hunk 11]
```diff
@@ -35,6 +35,8 @@
     KMM,
     ReweightDensityAdapter,
     ReweightDensity,
+    NearestNeighborDensityAdapter,
```
rflamary, line 38:
```suggestion
    NearestNeighborReweightDensityAdapter,
```

skada/__init__.py
```diff
@@ -89,6 +91,8 @@
     "KMM",
     "ReweightDensityAdapter",
     "ReweightDensity",
+    "NearestNeighborDensityAdapter",
```
rflamary, line 94: same here

rflamary (commented):
Hello this is very nice (especially the example comparing the different reweighting methods). But when looking at it I really believe we Need to change the Methods name and this would be the good PR to do this.

basically I think we do not need the density in the reweighting methods especially sine many of them do not estimate a density at all (except ReweightDensity).

Basically Here are the name I sugest for the methods (and Adapter by adding Adaper at the end)

```python
DiscriminatorReweight,
GaussianReweight,
DensityReweight, # the only one using a density estimator but with reweight at the end
NearestNeighborReweight
```

rflamary:
Hello @BuenoRuben sory for this infinitre PR ;) could you also rename `KMMAdapter ` to `KMMReweightAdapter` and same fro KLIEP ?

After that it seems everything will be OK

BuenoRuben:
should I then rename `KMM` into `KMMReweight` then?

PR #123, opened 2024-02-29
Remove masked inputs only if estimator does not accept `sample_domain`

kachayev:
Per the discussion in Slack.

`Selector` now removes masked inputs if the base estimator does not accept `sample_domain` through the routing (e.g. any sklearn estimator out there). The contract is that if the estimator declares `sample_domain` as a parameter, it should be able to deal with inputs being partially masked (presumably targets).


rflamary (approved):
This is very nice

PR #124, opened 2024-03-01
[MRG] Implement CircularValidation as a scorer

### LINKED ISSUE #111: Implement the Circular Validation Strategy
See [5]: https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=4803844
Implement the Circular Validation Strategy as a way to score a DA method.
This scorer should work with the `sklearn.model_selection.GridSearchCV`

YanisLalou:
Fixes #111 
Paper: https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=4803844&tag=1
Process described in part [5] and [6].

Small thing not in the paper but implemented:
- If the estimator predicts always the same value, then we assume that the `backward_estimator` will also predicts  always the same value.

--> Why? Because for some estimators, if you train them with np.unique(y) = 1 you get ValueErrors (happens with SVC).

### DIFF
--- a/skada/metrics.py
+++ b/skada/metrics.py
@@ -427,7 +427,13 @@ class DeepEmbeddedValidation(_BaseDomainAwareScorer):
 class CircularValidation(_BaseDomainAwareScorer):
     """Score based on a circular validation strategy.
 
-    See [1]_ for details.
+    This scorer retrain the estimator, with the exact same parameters,
+    on the predicted target domain samples. Then, the retrained estimator
+    is used to predict the source domain labels. The score is then
+    computed using the source_scorer between the true source labels and
+    the predicted source labels.
+
+    See [21]_ for details.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


skada/metrics.py
```diff
@@ -413,3 +422,105 @@ def _score(self, estimator, X, y, sample_domain=None, **kwargs):
         var_w = np.var(weights, ddof=1)
         eta = -cov / var_w
         return self._sign * (np.mean(weighted_error) + eta * np.mean(weights) - eta)
+
+
+class CircularValidation(_BaseDomainAwareScorer):
+    """Score based on a circular validation strategy.
+
+    See [1]_ for details.
```
rflamary, line 430: Bad number : use [21] as in teh redame

rflamary: Also small sentence to explain that this estimator retrain an estimator for each call this should be in the descriptin

skada/metrics.py
```diff
@@ -413,3 +422,105 @@ def _score(self, estimator, X, y, sample_domain=None, **kwargs):
...
+    ----------
+    .. [1]  Bruzzone, Lorenzo & Marconcini, Mattia.
+            Domain Adaptation Problems: A DASVM Classification Technique
+            and a Circular Validation Strategy. IEEE, 2010.
+    """
+
+    def __init__(
+        self,
+        source_scorer=balanced_accuracy_score,
+        random_state=None,
```
rflamary, line 452: remove random state if not used

rflamary (commented):
few comments

PR #125, opened 2024-03-01
[TO_REVIEW] _find_y_type return enum

### LINKED ISSUE #85: `_find_y_type` to return enum
Enums are more convenient to work with as they provide access to static and dynamic code analysis.

YanisLalou:
Fixes #85


kachayev:
Cool! The only question I have: should it be 'continuous' vs. 'discrete'? Or 'classification' vs. 'regression' (though it is not exactly a 'type of y', more like type of a task).

kachayev (approved):

PR #126, opened 2024-03-01
Add frank-wolfe algorithm

antoinedemathelin:
Hi everyone,
I propose to add the Frank-wolfe algorithm to speed KMM. (almost 20 times faster in the small example).
It can be used for KLIEP and Tars too, in the future.

Cf paper https://webdocs.cs.ualberta.ca/~dale/papers/ijcai15.pdf

Cf PR #93 PR #108

### DIFF
--- a/examples/methods/plot_kmm_reweighting.py
+++ b/examples/methods/plot_kmm_reweighting.py
@@ -13,6 +13,8 @@ This example illustrates the use of KMM method [1] to correct covariate-shift.
 # sphinx_gallery_thumbnail_number = 1
 
 # %%
+import time
+
 import matplotlib.pyplot as plt
 from matplotlib.colors import ListedColormap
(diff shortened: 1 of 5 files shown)


rflamary:
Hello @antoinedemathelin this is great.

I was thinking that we would  add a parameter to qpsolve and insteda have _qpsolve_scipy  and _qpsolve_frank_wolfe functions ? this would be nice because it means that in the methdos one can just apss the solver parameter.

rflamary:
I see that your solver is less general but one can use the general function and taise an error if the constrainst cannot be solve by FW

antoinedemathelin:
Hi @rflamary,
Ok I can do that!

rflamary:
I'm closing this PR since we have the new and improved #167

PR #130, opened 2024-03-03
Precommit with ruff+codespell

agramfort:
setting up precommit hooks with ruff + codespell

codespell found a few typos :)

feedback welcome


rflamary:
@agramfort thanks for the PR, tests are not running anymore since last commit. 
https://github.com/scikit-adaptation/skada/actions/runs/8133112804

Could we have a quick test checking that style is valid (flake8 was not needed anymore but can we test it with ruff maybe?).  To detect new contributors that did not install the pre-commit.

agramfort:
@rflamary @kachayev @tgnassou this one is good to go from my end

kachayev:
@agramfort This is cool, very needed tools! I have a question regarding testing/linting flow: testing job is setup to run pre-commit hooks, which contains ruff linter with `--fix` flag. What happens if someone commits unlinted file? I guess it will just fix the code locally in the test container without reporting/failing? 🤔

agramfort:
@kachayev good question. I copied the practice of another repo that has been using this successfully for a few months. Let's see how it goes here. We can always adjust.

rflamary:
well now we know it does not fail since my PR definitely does not respect ruff:
https://github.com/scikit-adaptation/skada/pull/135

kachayev:
That's interesting. In my PRs it actually fails because of files being changed (which is desirable) 🤔 

```
ruff lint skada..........................................................Failed
- hook id: ruff
- files were modified by this hook
- ```

rflamary:
this is very weird no?

PR #133, opened 2024-03-04
Ignore deep/* tests when torch is not installed

### LINKED ISSUE #132: Avoid the `--ignore skada/deep` in pytest CI
For now to skip the test for the deep methods we use `--ignore skada/deep` option in pytest. We can try to avoid that by using `importorskip` function of pytest, or moving the tests outside the module skada. The first is the best one but brought some problems when I tried it.

kachayev:
Closes #132


kachayev:
@tgnassou Seems to be working now. Would you please double check?

rflamary (approved):

Issue #96, opened 2024-02-20
Multidomain source target split and merge


YanisLalou:
Add the ability to have:
- `source_target_merge` from utils.py to accept `*List[arrays]` with each list corresponding to one sample domain. Ex: List[X_domain_1, y_domain_1], List[X_domain_2, y_domain_2] ...
- `source_target_split` from utils.py to return `*List[arrays]` with each list corresponding to one sample domain.

kachayev:
It sounds like a functionality of `DomainAwareDataset`, right?

rflamary:
Not totally, if you have access to `X,y,sample_domain` but want to plot individual domains (or estimate their) or the reverse sucjh functions make sens but now that I see it probably it is better to have `domain_merge`and `domain_split`function that can also return/take a list of domain numbers corresponding to the arrays in the lists.  For the moment we have no multidomain functiosn but it will be necessary to have such helpers inside the classes also (going back to DomainAwareDataset and doing a list of pack is not friendly for this )

PR #137, opened 2024-03-05
 [WIP] ConditionalTransferableComponentsAdapter Implementation

YanisLalou:
Fixes: #96 
Paper: https://proceedings.mlr.press/v48/gong16.pdf
Mostly eq 7-8-9-10

### DIFF
--- a/skada/_subspace.py
+++ b/skada/_subspace.py
@@ -475,6 +475,9 @@ class ConditionalTransferableComponentsAdapter(BaseAdapter):
         #TODO: Add constraints on alpha, W, G, H
         for i in range(self.max_iter):
             if i % 3 == 0:
+                # For α, we use quadratic programming (QP) to minimize
+                # Jˆct w.r.t. α under constraint (11)
+                #TODO: Switch to qp_solver here
                 alpha, _ = torch_minimize(func, alpha, args=(G, H, W), tol=self.tol, max_iter=1)
             elif i % 3 == 1:
                 W, _ = torch_minimize(func, W, args=(alpha, G, H) tol=self.tol, max_iter=1)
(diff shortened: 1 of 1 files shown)


skada/_subspace.py
```diff
@@ -1021,3 +1025,341 @@ def TransferSubspaceLearning(
...
+
+                # grassmann_constraint = {
+                #     "type": "eq",
+                #     "fun": grassmann_constraint_function,
+                # }
+
+                (_, W, _, _), _ = torch_minimize(
+                    func_torch, (alpha, W, G, H), tol=self.tol, max_iter=1
+                )
+                W = torch.tensor(W, dtype=torch.float64)
```
YanisLalou, line 1241: Here any idea how to enforce the constraint to the torch minimiser ?

ambroiseodt: hello, here are a couple of ideas:
- Use the dual of the optimization problem resulting in adding the constraint to the loss in Pytorch:
min_x  f(x) st g(x) = 0 <==> min_{x, lambda} f(x) + lambda * g(x)
here the constraint would be the norm of W^TW - Id

- Another idea could be to generate the matrices directly on the Grassmanian manifold:
 a) geomtorch: https://geotorch.readthedocs.io/en/latest/orthogonal/grassmannian.html 

b) torch: https://pytorch.org/docs/stable/generated/torch.nn.utils.parametrizations.orthogonal.html
Une autre methode, c'est de parametriser tes fonctions tq ce soit orthogonal avec 

- Maybe not directly related, but here is an interesting paper with code to optimize orthogonal matrices: 
https://proceedings.mlr.press/v151/ablin22a/ablin22a.pdf (code: https://github.com/pierreablin/landing)

YanisLalou:
By guaranteeing that W is on the Grassmann manifold we loose performance in the `Subspace method example on subspace shift dataset` as illustrated here:

With constraint:
image

Without constraint:
image

Is it expected or is there an issue ?

PR #140, opened 2024-03-05
[MRG] Implementation of the TJM method

BuenoRuben:
see: https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=a279f53f386ac78345b67e13c1808880c718efdf#page=3&zoom=100,144,841


skada/_subspace.py
```diff
@@ -297,13 +299,11 @@ def adapt(self, X, y=None, sample_domain=None, **kwargs):
...
             The data transformed to the target subspace.
-        y_t : array-like, shape (n_samples,)
-            The labels (same as y).
         sample_domain : array-like, shape (n_samples,)
             The domain labels transformed to the target subspace
             (same as sample_domain).
-        weights : None
-            No weights are returned here.
+        weigts : None
+            No weigts are returned here.
```
agramfort, line 306: seems like typos came back :)

skada/__init__.py
```diff
@@ -96,6 +98,8 @@
     "SubspaceAlignment",
     "TransferComponentAnalysisAdapter",
     "TransferComponentAnalysis",
+    "TJMAdapter",
+    "TJM",
```
agramfort, line 102: could we avoid acronyms for estimator names? TJM -> TransferJointMatching

skada/_subspace.py
```diff
@@ -271,7 +273,7 @@ def fit(self, X, y=None, sample_domain=None, **kwargs):
         B = K @ H @ K
         solution = np.linalg.solve(A, B)
 
-        eigvals, eigvects = np.linalg.eigh(solution)
+        eigvals, eigvects = np.linalg.eig(solution)
```
agramfort, line 276: this means that evals can be complex. Is that a problem?

skada/_subspace.py
```diff
@@ -366,3 +366,233 @@ def TransferComponentAnalysis(
         ),
         base_estimator,
     )
+
+
+class TJMAdapter(BaseAdapter):
+    """Domain Adaptation Using TJM: Transfer Joint .
```
agramfort, line 372:
```suggestion
    """Domain Adaptation Using TJM: Transfer Joint Matching.
```

skada/_subspace.py
```diff
@@ -366,3 +366,233 @@ def TransferComponentAnalysis(
...
+        -------
+        X_t : array-like, shape (n_samples, n_components)
+            The data transformed to the target subspace.
+        y_t : array-like, shape (n_samples,)
+            The labels (same as y).
+        sample_domain : array-like, shape (n_samples,)
+            The domain labels transformed to the target subspace
+            (same as sample_domain).
+        weigts : None
+            No weigts are returned here.
```
agramfort, line 440: these return statements seem wrong

skada/_subspace.py
```diff
@@ -366,3 +366,233 @@ def TransferComponentAnalysis(
...
+            self.A_ = A
+            l21 = self.l21_norm(A[:ns])
+            f = self.frobenius_norm(A[ns:])
+            c = np.trace(A.T @ K @ M @ K @ A) + self.regularizer * (l21 + f**2)
+            print(c)
+
+        return self
+
+
+def TJM(
```
agramfort, line 552: do we really need a function for this? this seems redundant with the estimator class

agramfort (commented):
precommit run -a

runs for you?

skada/_subspace.py
BuenoRuben: yes, you are right, I'm correcting that right now

skada/_subspace.py
BuenoRuben: yes, I just placed then so that pre-commit was not preventing me from committing, I had a problem with the algorithm and wanted @tgnassou to be able to see the algorithm, I will correct those now

skada/_subspace.py
BuenoRuben: ho, this is a mistake, I shouldn't be changing this

skada/_subspace.py
```diff
@@ -366,3 +365,240 @@ def TransferComponentAnalysis(
...
+            n_components = self.n_components
+        self.random_state_ = check_random_state(self.random_state)
+
+        n = X.shape[0]
+        H = np.identity(n) - 1 / n * np.ones((n, n))
+        M = self._get_mmd_matrix(X_source.shape[0], X_target.shape[0], sample_domain)
+        self.X_source_ = X_source
+        self.X_target_ = X_target
+        K = self._get_kernel_matrix(X_source, X_target)
+        M /= np.linalg.norm(M, ord="fr" "o")
```
BuenoRuben, line 523: is there a way for me to write "fro" without pre-commit raising an error? (because this is pretty ugly...)

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
+"""
+Subspace method example on covariate shift dataset
```
tgnassou, line 2: IT is not ht ecovariate shift dataset

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
...
+
+clf = SubspaceAlignment(base_classifier, n_components=1)
+clf.fit(X, y, sample_domain=sample_domain)
+create_plots(clf, "Subspace Alignment")
+
+# %%
+#     Illustration of the Transfer Component Analysis method
+# ------------------------------------------
+#
+# The goal of Transfer Component Analysis (or TCA) is to learn some transfer
```
tgnassou, line 212: just (TCA) remove the or

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
...
+# components across domains in a reproducing kernel Hilbert space using maximum
+# mean discrepancy (mmd).
+#
```
tgnassou, line 215: Maximum Mean Discrepancy (MMD)

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
...
+clf.fit(X, y, sample_domain=sample_domain)
+create_plots(clf, "TCA")
+
+# %%
+#     Illustration of the Transfer Joint Matching method
+# ------------------------------------------
+#
+# In most of the previous works, we explored two learning strategies independently for
+# domain adaptation: feature matching and instance reweighting. Transfer Joint Matching
+# or TJM aims to use both, by adding a constant to tradeoff between to two.
```
tgnassou, line 232: (TJM)

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
...
+    scores_dict[name] = score
+
+    ax = axes[0]
+
+    # Plot the source points:
+    ax.scatter(Xs[:, 0], Xs[:, 1], c=ys, cmap=colormap, alpha=0.7, s=size)
+
+    ax.set_xticks(()), ax.set_yticks(())
+    ax.set_xlim(x_min, x_max), ax.set_ylim(y_min, y_max)
+    ax.set_title("Training with rewegihted data", fontsize=12)
```
tgnassou, line 162: IT is no more reweighted method :)

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,264 @@
...
+    if name != "Without DA":
+        ax = axes[2]
```
tgnassou, line 164: pMaybe it could be nice to see how to get the subspace, and not hide it in the function

antoinecollas:
Now the constraints are always verified and the loss decreases when `tradeoff=0`.
However, when tradeoff is different from 0, the loss oscillates... So there is still a problem.

skada/tests/test_subspace.py
```diff
@@ -30,6 +32,10 @@
             TransferComponentAnalysisAdapter(n_components=2), LogisticRegression()
         ),
         TransferComponentAnalysis(n_components=2),
+        TransferJointMatching(n_components=2),
+        make_da_pipeline(
+            TransferJointMatchingAdapter(n_components=2), LogisticRegression()
+        ),
```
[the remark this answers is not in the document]
BuenoRuben: now by default it is 1, so I will keep the 2

BuenoRuben:
I managed to make the loss decrease when we have a tradeoff, but the conmition of one of the matrix being identity is still not solved, however the decision boundaries are closer to those of TCA, so we should be fine, however I don't understand why before, when the loss was oscillating, we had a better adapter (we had 100% acc)

skada/_subspace.py
```diff
@@ -366,3 +373,265 @@ def TransferComponentAnalysis(
...
+        # s.t. A^T K H K^T A = I
+        EPS_eigval = 1e-12
+        for i in range(self.max_iter):
+            # update A
+            B = K @ M @ K.T + self.tradeoff * G
+            B = B + EPS_eigval * np.identity(n)
+            C = K @ H @ K.T
+            C = C + EPS_eigval * np.identity(n)
+            solution = scipy.linalg.solve(B, C)
+            phi, A = scipy.linalg.eigh(solution)
```
antoinecollas, line 545: `scipy.linalg.eigh(B, C)` solves `B X = C X D` with `X` and `D` the eigenvectors and eigenvalues respectively. So `phi, A = scipy.linalg.eigh(B, C)`  is better.

skada/_subspace.py
```diff
@@ -366,3 +373,265 @@ def TransferComponentAnalysis(
...
+            phi = phi + EPS_eigval
+            indices = np.argsort(-np.abs(phi))[:n_components]
```
antoinecollas, line 547: `B` and `C` are symmetric (semi-) positive definite. So eigenvalues `phi` must be positive. So you have to remove `np.abs`, otherwise there is a problem somewhere.

You put a minus, so get the greatest eigenvalues. If I remember correctly, the algorithm requires the lowest one.

skada/_subspace.py
```diff
@@ -366,3 +373,265 @@ def TransferComponentAnalysis(
...
+                    f"iter {i}: loss={loss_total:.4f}, loss_mmd={loss:.4f}, "
+                    f"reg={reg:.4f}"
+                )
+                mat = A.T @ K @ H @ K.T @ A
+                cond = np.allclose(mat, np.identity(n_components))
+                print(
+                    f"Constraint satisfaction: {cond}, dist="
+                    f"{np.linalg.norm(mat - np.identity(n_components))}"
+                )
+                print(mat)
```
antoinecollas, line 575: In general, you don't want the user to see big matrices. So this line would be removed.

antoinecollas (commented):
Be careful with smallest of greatest eigenvalues !!

skada/_subspace.py
BuenoRuben: ok, I will do that

skada/_subspace.py
BuenoRuben: I don'tw know why and how, but if I don't pass by scipy.linalg.solve (as it was done in TCA), I get really weird (I'm sending you the image) when there is tradeoff, but exactly what we want (the constraint is what we want)
Figure_5
What is exactly the difference between using solve and not using it?

BuenoRuben: (don't worry about answering now, I have plenty of other things to do)

antoinecollas: Solving $B X = C X D$ or $C^{-1} B X = X D$ with respect $X$ and $D$ is the same. Have you checked the eigenvalues of $B$ and $C$ are positive (`np.linalg.eigvalsh(B)` and `np.linalg.eigvalsh(C)`)?

antoinecollas: Does the loss equal to `0` and are the contraints respected when `tradeoff = 0`?

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,347 @@
...
+)
+
+clf = TransferJointMatching(
+    base_classifier, kernel="rbf", tradeoff=0.1, n_components=1, max_iter=20, tol=0.1
+)
+clf.fit(X, y, sample_domain=sample_domain)
+create_plots(
+    clf,
+    "TransferJointMatching with rbf kernel",
+)
```
antoinecollas, line 328: You should put the same kernel as in the other methods, i.e. `rbf`, and remove `linear`.

antoinecollas: Also, let the default parameters (except `n_components`).

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+                print(
+                    f"iter {i}: loss={loss_total:.4f}, loss_mmd={loss:.4f}, "
+                    f"reg={reg:.4f}"
+                )
+                mat = A.T @ K @ H @ K.T @ A
+                cond = np.allclose(mat, np.identity(n_components))
+                print(
+                    f"Constraint satisfaction: {cond}, dist="
+                    f"{np.linalg.norm(mat - np.identity(n_components))}"
+                )
```
antoinecollas, line 583: These lines are not tested. A test with `verbose=True` must be added.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+        M = self._get_mmd_matrix(X_source.shape[0], X_target.shape[0], sample_domain)
+        M /= np.linalg.norm(M, ord="fro")
+        G = np.identity(n)
+
+        # minimization of the objective function
+        # \min_{A} tr(A^T K M K A) + tradeoff * (||A_s||_{2, 1} + ||A_t||_F^2)
+        # s.t. A^T K H K^T A = I
+        EPS_eigval = 1e-10
+        # We take something that will force at least one iteration since loss>=0:
+        last_loss = -2 * self.tol
```
antoinecollas, line 544: Remove the comment and put `last_loss = np.inf`.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+                )
+                mat = A.T @ K @ H @ K.T @ A
+                cond = np.allclose(mat, np.identity(n_components))
+                print(
+                    f"Constraint satisfaction: {cond}, dist="
+                    f"{np.linalg.norm(mat - np.identity(n_components))}"
+                )
+
+            if np.abs(last_loss - loss_total) < self.tol:
+                break
```
antoinecollas, line 586: The criteria must be relative, e.g `np.abs((last_loss - loss_total) / last_loss)`.

skada/tests/test_subspace.py
```diff
@@ -30,6 +32,7 @@
             TransferComponentAnalysisAdapter(n_components=2), LogisticRegression()
         ),
         TransferComponentAnalysis(n_components=2),
+        TransferJointMatching(n_components=1, tradeoff=0.1),
```
antoinecollas, line 35: You must add a `make_da_pipeline` with TJM.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+
+        self.A_ = A
+
+        return self
+
+
+def TransferJointMatching(
+    base_estimator=None,
+    random_state=None,
+    n_components=1,
```
antoinecollas, line 598: The default `n_components` must be consistent with the other methods, i.e. `None`.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+            A_norms = np.linalg.norm(A, axis=1)
+            G = np.zeros(n, dtype=np.float64)
+            # ||A_s||_{2, 1}
+            G[A_norms != 0] = 1 / (2 * A_norms[A_norms != 0] + EPS_eigval)
+            # ||A_t||_F^2
+            G[~source_mask] = 1
+            G = np.diag(G)
+
+            loss = np.trace(A.T @ K @ M @ K @ A)
+            # np.linalg.norm(A[source_mask], axis=1).sum()
```
antoinecollas, line 566: This comment must be removed.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+    ----------
+    None
+
+    References
+    ----------
+    .. [1]
+            [Long et al., 2014] Long, M., Wang, J., Ding, G., Sun, J., and Yu, P.
+            (2014). Transfer joint matching for unsupervised domain adaptation.
+            In IEEE Conference on Computer Vision and Pattern Recognition (CVPR),
+            pages 1410–1417
```
antoinecollas, line 416: It should be `[1] Long, M., Wang .......`.

skada/_subspace.py
```diff
@@ -366,3 +373,281 @@ def TransferComponentAnalysis(
...
+
+    """
+
+    def __init__(
+        self,
+        n_components=None,
+        random_state=None,
```
antoinecollas, line 423: Why do we need a `random_state`?

examples/methods/plot_subspace.py
BuenoRuben: the default parametter means no tradeoff, it would be a shame as the whole point of thie method is to be able to have this tradeoff, but for the rest, I'll do as you said

skada/tests/test_subspace.py
```diff
@@ -52,6 +59,9 @@ def test_subspace_alignment(estimator, da_dataset):
         (TransferComponentAnalysisAdapter(), 5, 3, 3),
         (TransferComponentAnalysisAdapter(), 2, 3, 3),
         (TransferComponentAnalysisAdapter(), 2, 5, 4),
+        (TransferJointMatchingAdapter(), 5, 3, 3),
```
tgnassou, line 62: This test is to try if when we use `n_component=None` we use the good value, so for you, it is ` n_components = min(min(X_source.shape), min(X_target.shape))`.

skada/_subspace.py
```diff
@@ -366,3 +375,277 @@ def TransferComponentAnalysis(
...
+        """
+        X, sample_domain = check_X_domain(
+            X,
+            sample_domain,
+            allow_multi_source=True,
+            allow_multi_target=True,
+        )
+        X_source, X_target = source_target_split(X, sample_domain=sample_domain)
+
+        if self.n_components is None:
```
tgnassou, line 523: Can you change the choice of `n_component` for the SA method? It is not the same as TCA. Except it was wanted @kachayev

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,336 @@
...
+#
+# When not using domain adaptation, the classifier won't train on
+# data that is distributed as the target sample domain, it will thus
+# not be performing optimaly.
+
+# We create a dict to store scores:
+scores_dict = {}
+
+
+def create_plots(
```
tgnassou, line 113: `plot_subspace_and_classifier`

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,336 @@
...
+
+        ax.set_xticks(()), ax.set_yticks(())
+        ax.set_ylim(-50, 50)
+        ax.set_title("Full dataset projected on the subspace", fontsize=12)
+
+        ax = axes[1]
+        Xt_adapted = subspace_estimator.adapt(Xt)
+        ax.scatter(
+            Xt_adapted,
+            [0] * Xt.shape[0],
```
tgnassou, line 181: Don't put label color on target for the first plot

examples/methods/plot_subspace.py
```diff
@@ -0,0 +1,336 @@
...
+            alpha=0.02,
+            s=size,
+        )
+        ax.set_ylim(-50, 50)
+        ax.set_title("Accuracy on target projected on the subspace", fontsize=12)
+    else:
+        ax = axes[0]
+        # Plot the source points:
+        X_subspace = X
+        ax.scatter(
```
tgnassou, line 204: Same here no color for target in the first plot

skada/_subspace.py
```diff
@@ -366,3 +375,277 @@ def TransferComponentAnalysis(
...
+                last_loss = loss_total
+
+        self.A_ = A
+
+        return self
+
+
+def TransferJointMatching(
+    base_estimator=None,
+    random_state=None,
```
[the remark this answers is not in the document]
BuenoRuben: I did I think I just didn't pushed it... I ll dmo it then

skada/tests/test_subspace.py
BuenoRuben: oui, donc 3 ici, non? (n_sample>n_feature), et on a 5 et 3 comme features, donc c'est 3

examples/methods/plot_subspace.py
BuenoRuben: sorry, I'm not sure I understood what you want me to do:
I should remove the color for label datapoints in the `Full dataset projected on the subspace` plot?
(and thus add a legend to explain this)

skada/tests/test_subspace.py
```diff
@@ -48,10 +55,13 @@ def test_subspace_alignment(estimator, da_dataset):
...
     [
         (SubspaceAlignmentAdapter(), 5, 3, 3),
-        (SubspaceAlignmentAdapter(), 2, 5, 2),
+        (SubspaceAlignmentAdapter(), 3, 3, 3),
         (TransferComponentAnalysisAdapter(), 5, 3, 3),
         (TransferComponentAnalysisAdapter(), 2, 3, 3),
         (TransferComponentAnalysisAdapter(), 2, 5, 4),
+        (TransferJointMatchingAdapter(), 5, 3, 3),
+        (TransferJointMatchingAdapter(), 2, 3, 3),
+        (TransferJointMatchingAdapter(n_components=4), 2, 5, 4),
```
tgnassou, line 64: This test is made to check that the right `n_components` is taken for `n_component=None`.

PR #141, opened 2024-03-07
[WIP] Geodesic Flow Kernel

antoinecollas:
This PR is a WIP implementation of [Geodesic Flow Kernel](https://projet.liris.cnrs.fr/imagine/pub/proceedings/CVPR2012/data/papers/261_O2D-03.pdf).

This method requires the [Generalized SVD](https://en.wikipedia.org/wiki/Generalized_singular_value_decomposition), and it is not implemented either in numpy or Scipy.
The `transferlearning` library uses `bob.math.gsvd` [here](https://github.com/jindongwang/transferlearning/blob/d740c84f8d5b8b42742b8429cd52c726b1e84c20/code/traditional/GFK/GFK.py#L58), which has its custom C++ code [here](https://github.com/bioidiap/bob.math/blob/master/bob/math/gsvd.cpp).

### DIFF
--- a/skada/__init__.py
+++ b/skada/__init__.py
@@ -23,6 +23,7 @@ from ._mapping import (
     MMDLSConSMapping,
     OTMappingAdapter,
     OTMapping,
+    GFKAdapter
 )
 from ._reweight import (
     DiscriminatorReweightDensityAdapter,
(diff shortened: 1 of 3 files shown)


antoinecollas:
No, I tried but the gsvd of the paper is hard to implement in python right now.

PR #149, opened 2024-03-11
[MRG] Propagate adaptation output through multiple steps

### LINKED ISSUE #146: Propogate `sample_weight` for resampling adapters through multiple steps
With the current implementation of how `AdaptationOutput` is propagated through the pipeline we should have troubles when reweighting adapter (producer of weights) and the estimator (consumer) are separated, e.g. when there's another transformer in between of them.

kachayev:
Closes #146


kachayev:
@rflamary This one seems to be ready. The processing gets more and more complex as we have a wider range of selectors to deal with. The code needs to be refactored to move common logical blocks out (to utils) though a specific way of doing so will definitely require some time and exploration. I will gladly tackle it in a separate PR.

rflamary (approved):

PR #156, opened 2024-03-25
[MRG] Fix Tars & MMDSConS

YanisLalou:
The R matrix is supposed to have a shape of (m, c) with m the len of X_source and c the len of unique y_source.
Otherwise we would have ValueErrors line 774 and 1274: `R[:, i] = (y_source == c).int()` when `X_target.shape[0] != len(y_source)`


skada/_mapping.py
```diff
@@ -769,7 +769,7 @@ def _mapping_optimization(self, X_source, X_target, y_source):
         # compute R
         if discrete:
             self.classes_ = classes = torch.unique(y_source).numpy()
-            R = torch.zeros((X_target.shape[0], len(classes)), dtype=torch.float64)
+            R = torch.zeros((X_source.shape[0], len(classes)), dtype=torch.float64)
```
antoinecollas, line 772: Can you replace `X_source.shape[0]` by `m` which is already defined?

skada/_reweight.py
```diff
@@ -1269,7 +1269,7 @@ def _weights_optimization(self, X_source, X_target, y_source):
         # compute R
         if discrete:
             self.classes_ = classes = np.unique(y_source)
-            R = np.zeros((X_target.shape[0], len(classes)))
+            R = np.zeros((X_source.shape[0], len(classes)))
```
antoinecollas, line 1272: Same here, replace `X_source.shape[0]` by `m` which is already defined.

antoinecollas (commented):
Nice catch 👍🏻 
But it also means that in the tests `X_target.shape[0] == X_source.shape[0]`. Can you change a test so that we test this case?

skada/tests/test_mapping.py
```diff
@@ -192,3 +192,54 @@ def test_new_X_adapt(estimator, da_dataset):
...
+        return_dataset=False,
+    )
+
+    X_source, X_target, y_source, y_target = source_target_split(
+        X, y, sample_domain=sample_domain
+    )
+    assert X_source.shape[0] < X_target.shape[0]
+
+    # No exception should be raised
+    estimator.fit(X, y, sample_domain=sample_domain)
```
antoinecollas, line 245: Adding new tests is maybe too much. Can you find an already existing test and change the data generation so that `X_source.shape[0] != X_target.shape[0]`?

rflamary: agreed

antoinecollas:
Changing somewhere `make_shifted_datasets(n_samples_source, n_samples_target, ...)` with `n_samples_source != n_samples_target` in the `contest.py` should be enough no?

YanisLalou:
> Changing somewhere `make_shifted_datasets(n_samples_source, n_samples_target, ...)` with `n_samples_source != n_samples_target` in the `contest.py` should be enough no?

It was already the case i.e `n_samples_source=20, n_samples_target=21`. Also in that case I can't test when `X_source.shape[0] < X_target.shape[0]` AND `X_source.shape[0] > X_target.shape[0]`

rflamary:
If you want to etst different sizes you shoulh parametrize the datasets instead of doing multiple tests. This is indeed an imprtant test when the two sizes are different and larger/smaller.

antoinecollas:
An it is strange that `n_samples_source != n_samples_target` and the tests pass???

skada/tests/test_mapping.py
```diff
@@ -157,22 +157,48 @@ def _base_test_new_X_adapt(estimator, da_dataset):
...
+def test_new_X_adapt(estimator, da_dataset, n_samples_source, n_samples_target):
+    da_dataset = make_shifted_datasets(
+        n_samples_source=n_samples_source,
+        n_samples_target=n_samples_target,
+        shift="concept_drift",
+        mean=0.5,
+        noise=0.3,
+        label="regression",
+        random_state=42,
+    )
```
antoinecollas, line 201: You throw away `da_dataset` and then generate a new dataset. IMO, you should modify the fixture in the `conftest.py` that outputs `da_dataset`, so that it outputs `da_datasets`, a list of datasets. Then, you do a for loop in the test. The overall idea is to generate data only in the `conftest.py`.

PR #160, opened 2024-04-03
[MRG] Add kwargs to DASVM predict/predict_proba + add score func

YanisLalou:
- Add kwargs to DASVM predict/predict_proba
- add score func (SupervisedScorer need it)

### DIFF
--- a/skada/_self_labeling.py
+++ b/skada/_self_labeling.py
@@ -263,7 +263,7 @@ class DASVMClassifier(DAEstimator):
         """Return predicted value by the fitted estimator for `X`
         `predict` method from the estimator we fitted
         """
-        return self.base_estimator_.predict(X)
+        return self.base_estimator_.predict(X, **kwargs)
 
     def _check_proba(self):
         if hasattr(self.base_estimator, "predict_proba"):
@@ -279,7 +279,7 @@ class DASVMClassifier(DAEstimator):
         `predict_proba` method from the estimator we fitted
         """
         check_is_fitted(self)
-        return self.base_estimator.predict_proba(X)
+        return self.base_estimator.predict_proba(X, **kwargs)
 
     def decision_function(self, X):
         """Return values of the decision function of the
@@ -296,4 +296,4 @@ class DASVMClassifier(DAEstimator):
                 "Source domain detected. Predictor is trained on target"
                 "and score might be biased."
             )
-        return self.base_estimator_.score(X, y, sample_weight=sample_weight)
+        return self.base_estimator_.score(X, y, sample_weight=sample_weight, **kwargs)
--- a/skada/_self_labeling.py
+++ b/skada/_self_labeling.py
@@ -53,6 +53,20 @@ class DASVMClassifier(DAEstimator):
             IEEE transactions on pattern analysis and machine intelligence, (2009).
     """
 
+    __metadata_request__fit = {"sample_domain": True}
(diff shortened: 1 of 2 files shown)


tgnassou (approved):

skada/_self_labeling.py
```diff
@@ -273,7 +274,7 @@ def _check_proba(self):
             )
 
     @available_if(_check_proba)
-    def predict_proba(self, X):
+    def predict_proba(self, X, **kwargs):
```
tgnassou, line 277: You add the `**kwargs` in the definition of the function but forgot to give them to `self.base_estimator.predict_proba(X, **kwargs)`

skada/_self_labeling.py
```diff
@@ -286,3 +287,13 @@ def decision_function(self, X):
...
+
+    def score(self, X, y, sample_domain=None, *, sample_weight=None, **kwargs):
+        """Return the scores of the prediction"""
+        check_is_fitted(self)
+        if sample_domain is not None and np.any(sample_domain >= 0):
+            warnings.warn(
+                "Source domain detected. Predictor is trained on target"
+                "and score might be biased."
+            )
+        return self.base_estimator_.score(X, y, sample_weight=sample_weight)
```
tgnassou, line 299: same here, missing `**kwargs`

skada/_self_labeling.py
```diff
@@ -258,7 +259,7 @@ def fit(self, X, y=None, sample_domain=None):
 
         return self
 
-    def predict(self, X):
+    def predict(self, X, **kwargs):
```
tgnassou, line 262: same problem as below

YanisLalou:
> You add the `**kwargs` in the definition of the function but forgot to give them to `self.base_estimator.predict_proba(X, **kwargs)`
By doing so we have `E TypeError: predict_proba() got an unexpected keyword argument 'sample_domain'` because BaseSVC doesnt accept ***kwargs.

2 possible solutions:
```
# If kwargs are accepted, pass them to predict_proba
        try:
            return predict_proba(X, **kwargs)
        except TypeError:
            # If kwargs are not accepted, call predict_proba without kwargs
            return predict_proba(X)
```
OR
```
# Check if 'kwargs' is among the parameters
    if 'kwargs' in predict_proba_signature.parameters:
        # If 'kwargs' is accepted, pass it to predict_proba
        return predict_proba(X, **kwargs)
    elif 'sample_domain' in predict_proba_signature.parameters:
        # If 'sample_domain' is accepted, pass it along with X
        return predict_proba(X, sample_domain)
    else:
        # If neither 'kwargs' nor 'sample_domain' is accepted, call predict_proba without them
        return predict_proba(X)
```

What do you think ?

YanisLalou:
OR

we can change the metadata for this specific object i.e:

```
    __metadata_request__fit = {'sample_domain': True}
    __metadata_request__partial_fit = {'sample_domain': False}
    __metadata_request__predict = {'sample_domain': False, 'allow_source': False}
    __metadata_request__predict_proba = {'sample_domain': False, 'allow_source': False}
    __metadata_request__predict_log_proba = {
        'sample_domain': False,
        'allow_source': False
    }
    __metadata_request__score = {'sample_domain': False, 'allow_source': False}
    __metadata_request__decision_function = {
        'sample_domain': False,
        'allow_source': False
    }
```

Issue #139, opened 2024-03-05
Deep copy a DomainAwareNet after fitting


YanisLalou:
After fitting a `DomainAwareNet`, it's not possible to deep copy it.
Strangely it works before the fitting.
Error raised:
`AttributeError: Can't pickle local object '_get_intermediate_layers.<locals>.hook'`

It's necessary to be able to deep copy it to use the `CircularValidation` scorer.

To reproduce:
````
import torch
import numpy as np
from skada.deep.base import DomainAwareCriterion, DomainAwareModule, DomainAwareNet, DomainBalancedDataLoader, BaseDALoss
from skada.metrics import (
    ImportanceWeightedScorer,
    PredictionEntropyScorer,
    SoftNeighborhoodDensity,
    DeepEmbeddedValidation,
    CircularValidation,
)
from skada.deep.modules import ToyModule2D
from skada import make_da_pipeline
from sklearn.model_selection import ShuffleSplit, cross_validate
from skada.datasets import make_shifted_datasets
from copy import deepcopy


class TestLoss(BaseDALoss):
    """Test Loss to check the deep API"""

    def __init__(
        self,
    ):
        super().__init__()

    def forward(
        self,
        *args,
    ):
        """Compute the domain adaptation loss"""
        return 0


da_dataset = make_shifted_datasets(
        n_samples_source=20,
        n_samples_target=20,
        shift="concept_drift",
        noise=0.1,
        random_state=42,
        return_dataset=True,
    )

X, y, sample_domain = da_dataset.pack_train(as_sources=["s"], as_targets=["t"])
X_test, y_test, sample_domain_test = da_dataset.pack_test(as_targets=["t"])

module = ToyModule2D()

estimator = DomainAwareNet(
	DomainAwareModule(module, "dropout"),
	iterator_train=DomainBalancedDataLoader,
	criterion=DomainAwareCriterion(torch.nn.CrossEntropyLoss(), TestLoss()),
	batch_size=10,
	max_epochs=2,
	train_split=None,
)

X = X.astype(np.float32)
X_test = X_test.astype(np.float32)

estimator_copy = deepcopy(estimator) # Doesn't raise errors

estimator.fit(X, y, sample_domain=sample_domain)

estimator_copy_after_fit = deepcopy(estimator) # Raises AttributeError
````

kachayev:
[...]

PR #169, opened 2024-04-23
[MRG] Fix issue of the circular validation with deep models

### LINKED ISSUE #163: CircularValidation does not work with deep method
To fix this problem, we must change how we do `deepcopy` of the estimator for deep methods.

YanisLalou:
Fixes #163 #139

### DIFF
--- a/skada/metrics.py
+++ b/skada/metrics.py
@@ -552,21 +552,17 @@ class CircularValidation(_BaseDomainAwareScorer):
         X, y, sample_domain = check_X_y_domain(X, y, sample_domain)
         source_idx = extract_source_indices(sample_domain)
 
-        # if estimator is NeuralNet
-        if not isinstance(estimator, BaseEstimator):
-            # The estimator is a deep model
-            from skorch import NeuralNet
-
-            if isinstance(estimator, NeuralNet):
-                backward_estimator = clone(estimator)
-            else:
-                raise ValueError(
-                    "The estimator passed should be a scikit-learn estimator or "
-                    "a skorch.NeuralNet "
-                    f"The estimator {estimator!r} is not."
-                )
-        else:
+        try:
             backward_estimator = deepcopy(estimator)
+        except (TypeError, AttributeError):
+            try:
+                backward_estimator = clone(estimator)
+            except Exception:
+                raise ValueError(
+                    "The estimator passed should be a scikit-learn estimator, "
+                    "a skorch.NeuralNet, or an object that supports deepcopy or clone. "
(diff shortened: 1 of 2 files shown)


skada/metrics.py
```diff
@@ -552,9 +552,25 @@ def _score(self, estimator, X, y, sample_domain=None):
...
+        if not isinstance(estimator, BaseEstimator):
+            # The estimator is a deep model
+            from skorch import NeuralNet
 
-        y_pred_target = estimator.predict(X[~source_idx])
+            if isinstance(estimator, NeuralNet):
+                backward_estimator = clone(estimator)
+            else:
+                raise ValueError(
+                    "The estimator passed should be a scikit-learn estimator or "
```
kachayev, line 564: Why restrict to either 2 of them? It seems like anything that supports `deepcopy` should be fine here, right?

If we do have our own wrapper for `skorch.NeuralNet`, we just need `__deepcopy__` method in there. If no, it would be better to check if the estimator is `skorch.NeuralNet` to make `clone` or `deepcopy` otherwise. If the user passes something that is not `NeuralNet` **and** it doesn't support `deepcopy`, they will get `copy.Error` (I guess).

skada/metrics.py
```diff
@@ -552,9 +552,21 @@ def _score(self, estimator, X, y, sample_domain=None):
...
-        backward_estimator = deepcopy(estimator)
-
-        y_pred_target = estimator.predict(X[~source_idx])
+        try:
+            backward_estimator = deepcopy(estimator)
+        except (TypeError, AttributeError):
+            try:
+                backward_estimator = clone(estimator)
+            except Exception:
+                raise ValueError(
```
rflamary, line 561: add a test here please that chack teh raise ?

kachayev:
In light of skorch Net not supporting deep copy, there are 2 possible scenarios:

1. It's just an oversight on their end. Would be nice to open issue or PR for them to explain the need, maybe they would just add __deepcopy__ on their side (so we will be able to remove workaround on our side).

2. There's a deeper non obvious issue that prevents Net objects from being "deep-copyable" in the first place (which could easily happen because Torch is mostly in C and C objects/initializers are not always done friendly for performing copy from Python side). It would be absolutely necessary for us to understand those limitations.

rflamary:
We need a working CircCV and we cannot wait fro skorch developers butw e shoudl definitely open an issue on skorch to ask about your point 1.

skada/metrics.py
kachayev: I would honestly remove this try/catch at all. The error message is somewhat misleading anyways.

PR #170, opened 2024-04-24
Allow multi dim inputs through the pipeline

kachayev:
Code sample from the Slack conversation:

```python
# %%
from pyriemann.estimation import Covariances
from pyriemann.tangentspace import TangentSpace
from skada import make_da_pipeline
from skada import CORALAdapter
import numpy as np
from skada.datasets import DomainAwareDataset
# %%
pipe = make_da_pipeline(
    Covariances(estimator='oas'),
    TangentSpace(),
    CORALAdapter()
)
# %%
Xs = np.random.randn(100, 22, 30)
Xt = np.random.randn(100, 22, 30)
ys = np.random.randint(0, 2, 100)
yt = np.random.randint(0, 2, 100)

# %%
dataset = DomainAwareDataset()
dataset.add_domain(Xs, ys, 'source')
dataset.add_domain(Xt, yt, 'target')
# %%
X, y, sample_domain = dataset.pack_train(as_sources=["source"], as_targets=["target"])
# %%
pipe.fit(X, y, sample_domain=sample_domain)
```

### DIFF
--- a/skada/tests/test_self_labeling.py
+++ b/skada/tests/test_self_labeling.py
@@ -66,6 +66,7 @@ def test_dasvm_estimator_predict_proba():
         shift="covariate_shift",
         noise=None,
         label="binary",
+        random_state=0,
     )
     X, y, sample_domain = check_X_y_domain(X, y, sample_domain)
 
--- a/skada/base.py
+++ b/skada/base.py
@@ -340,7 +340,8 @@ class BaseSelector(BaseEstimator, _DAMetadataRequesterMixin):
 
         X_out, sample_domain = check_X_domain(
             X_out,
-            sample_domain=params.get('sample_domain')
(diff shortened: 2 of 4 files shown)


skada/tests/test_self_labeling.py
```diff
@@ -66,6 +66,7 @@ def test_dasvm_estimator_predict_proba():
         shift="covariate_shift",
         noise=None,
         label="binary",
+        random_state=0,
```
rflamary, line 69:
```suggestion
        random_state=None,
```
?

rflamary: sorry I did not see this was   atest

kachayev: all good 👍 it would be likely better to use 42, which is common for other tests cases

PR #174, opened 2024-04-29
Make deep DA methods compatible with GPU

Florent-Michel:


skada/deep/base.py
```diff
@@ -253,17 +253,26 @@ def forward(self, X, sample_domain=None, is_fit=False, return_features=False):
...
+                )
                 domain_pred[source_idx] = domain_pred_s
                 domain_pred[~source_idx] = domain_pred_t
             else:
                 domain_pred = None
 
-            y_pred = torch.empty((len(sample_domain), y_pred_s.shape[1]))
+            y_pred = torch.empty(
+                (len(sample_domain), y_pred_s.shape[1]),
+                device = y_pred_s.device
```
agramfort, line 267: this is a pep8 violation. How come this is not caught by CIs?

Florent-Michel: I don't know about the CI, but I fixed it in another PR anyway, thanks.

PR #180, opened 2024-04-30
[TO-REVIEW] Add multi-domain Monge alignment and JCPOT Target shift method

rflamary:
Adds a multi domain Monge alignement Adapter to SKADA that does the following
1. compte a barycenter of all sources (Gaussian assumtion)
2. Map all data to this barycenter with the clase form
3. At test time it maps all data (form target or source) to the barycenter when calling tranform

Also implement the multisource JCPOT taregt shift method (that should also work for only one source)

Remain to be done
- [x] Proper documentation
- [x] pipeline function wit adapter
- [x] Add references
- [x] Implement JCPOT


examples/methods/plot_monge_alignment_da.py
```diff
@@ -0,0 +1,259 @@
...
+
+print(
+    "Average accuracy on all domains:",
+    clf.score(X, y, sample_domain=sample_domain, allow_source=True),
+)
+
+# %% Multisource and taregt data
+
+
+def get_multidomain_data(
```
kachayev, line 126: This should go to toy datasets, right?

skada/_mapping.py
```diff
@@ -500,6 +503,149 @@ def LinearOTMapping(
...
+        y : array-like, shape (n_samples,)
+            The source labels.
+        sample_domain : array-like, shape (n_samples,)
+            The domain labels.
+
+        Returns
+        -------
+        X_t : array-like, shape (n_samples, n_components)
+            The data transformed to the target subspace.
+        y_t : array-like, shape (n_samples,)
```
kachayev, line 630: Only `X_adapt` is returned.

skada/utils.py
```diff
@@ -299,6 +299,54 @@ def source_target_split(
...
+        dict of source domains (contain tuple of subset of arrays).
+    target_dict : dict
+        dict of target domains (contain tuple of subset of arrays).
+    """
+    if len(arrays) == 0:
+        raise ValueError("At least one array required as input")
+
+    check_consistent_length(arrays)
+
+    domain_idx = extract_domains_indices(sample_domain, False)
```
kachayev, line 331: We need to make boolean into a (forced-)named argument (I guess not the part of this PR)

PR #181, opened 2024-04-30
[MRG] Add TransferSubspaceLearning method

antoinecollas:
Add [TransferSubspaceLearning](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=4118b4fc7d61068b9b448fd499876d139baeec81) method

### DIFF
--- a/skada/_subspace.py
+++ b/skada/_subspace.py
@@ -652,7 +652,7 @@ class TransferSubspaceLearningAdapter(BaseAdapter):
     Parameters
     ----------
     n_components : int, default=None
-        The numbers of components to learn with PCA.
+        The numbers of components to learn.
         Should be less or equal to the number of samples
         of the source and target data.
     base_method : str, default='flda'
@@ -664,7 +664,6 @@ class TransferSubspaceLearningAdapter(BaseAdapter):
     reg : float, default=1e-4
         The regularization parameter of the covariance estimator.
         Possible values:
-
           - None: no shrinkage.
           - float between 0 and 1: fixed shrinkage parameter.
     max_iter : int>0, default=100
@@ -866,7 +865,10 @@ class TransferSubspaceLearningAdapter(BaseAdapter):
         y_source = torch.tensor(y_source)
         X_target = torch.tensor(X_target, dtype=torch.float64)
 
-        # Loss function
+        # Solve the optimization problem
+        # min_W F(W) + mu * D(W)
+        # s.t. W^T W = I
+
         def _orth(W):
             if type(W) is np.ndarray:
                 W = np.linalg.qr(W)[0]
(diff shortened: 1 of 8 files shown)


skada/_subspace.py
```diff
@@ -640,3 +642,322 @@ def TransferJointMatching(
...
+        sample_domain : array-like, shape (n_samples,)
+            The domain labels transformed to the target subspace
+            (same as sample_domain).
+        weights : None
+            No weights are returned here.
+        """
+        X, sample_domain = check_X_domain(
+            X,
+            sample_domain,
+            allow_multi_source=True,
```
rflamary, line 743:
```suggestion
            allow_multi_source=False,
```

skada/_subspace.py
```diff
@@ -640,3 +642,322 @@ def TransferJointMatching(
...
+        sigma_22 = 2 * sigma_2
+
+        L_11 = torch.linalg.cholesky(torch.linalg.inv(sigma_11))
+        L_12 = torch.linalg.cholesky(torch.linalg.inv(sigma_12))
+        L_22 = torch.linalg.cholesky(torch.linalg.inv(sigma_22))
+        Kss = torch.exp(-0.5 * torch.cdist(Z_source @ L_11, Z_source @ L_11))
+        Kst = torch.exp(-0.5 * torch.cdist(Z_source @ L_12, Z_target @ L_12))
+        Ktt = torch.exp(-0.5 * torch.cdist(Z_target @ L_22, Z_target @ L_22))
+
+        return torch.mean(Kss) + torch.mean(Ktt) - 2 * torch.mean(Kst)
```
YanisLalou, line 790: Why don't we return 2 times the value of _D? While it's true that we ultimately multiply it by mu later in the process, omitting this step might lead to confusion for readers accustomed to this method, as they would typically expect to use a specific mu value.

skada/_subspace.py
```diff
@@ -640,3 +642,322 @@ def TransferJointMatching(
...
+            The source labels.
+        sample_domain : array-like, shape (n_samples,)
+            The domain labels (same as sample_domain).
+
+        Returns
+        -------
+        self : object
+            Returns self.
+        """
+        try:
```
kachayev, line 836: I guess it is better to say so when instantiating the object (i.e. in the init) rather than when calling fit

skada/_subspace.py
```diff
@@ -640,3 +642,322 @@ def TransferJointMatching(
...
+        X_target = torch.tensor(X_target, dtype=torch.float64)
+
+        # Solve the optimization problem
+        # min_W F(W) + mu * D(W)
+        # s.t. W^T W = I
+
+        def _orth(W):
+            if type(W) is np.ndarray:
+                W = np.linalg.qr(W)[0]
+            else:
```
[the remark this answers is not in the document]
antoinecollas: `torch_minimize` outputs a `np.ndarray` !

skada/_subspace.py
```diff
@@ -640,3 +642,322 @@ def TransferJointMatching(
...
+            classes_means = classes_means * torch.sqrt(classes_n_samples).reshape(-1, 1)
+            S_W = self._torch_cov(classes_means)
+            S_B = self._torch_cov(X_source)
+            loss = torch.trace(W.T @ S_W @ W) / torch.trace(W.T @ S_B @ W)
+        elif base_method == "lpp":
+            # E is the Gaussian kernel if (y_source)_i == (y_source)_j and 0 otherwise
+            E = torch.exp(-0.5 * torch.cdist(X_source, X_source))
+            E = E * (y_source[:, None] == y_source[None, :])
+            D = torch.diag(torch.sum(E, dim=1))
+            loss = -torch.trace(W.T @ X_source.T @ (D - E) @ X_source @ W)
```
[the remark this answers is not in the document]
antoinecollas: Thanks for catching this!

skada/_subspace.py
antoinecollas: There is no factor 2 in equation (8). Did I miss something?

YanisLalou: No you're right, I mixed up eq 8 and eq 10

antoinecollas:
@kachayev I updated the PR. I kept the `import torch` in the `fit` method. Otherwise, you can't even instantiate the class and then I have no idea how to get the CI `test-minimal` working.
Also, we did similar things in other methods requiring `torch` for the optimization.

kachayev:
Yeah, I got it.

I think if we allow torch as an optional dependency, we should never even export classes that are non-functional without torch being installed (i.e. those classes should not be in skada namespace). Same way it works for deep modules. But I agree this is a separate issue as we already have a few places where this happens.

PR #183, opened 2024-04-30
[FIX][ENH] Fix and add more explanation to "DA validation procedures" examples

apmellot:
I made changes in the "Using cross_val_score with skada" example. I will now have a look at the two other examples of this section.

### DIFF
--- a/examples/validation/plot_cross_val_score_for_da.py
+++ b/examples/validation/plot_cross_val_score_for_da.py
@@ -2,11 +2,13 @@
 Using cross_val_score with skada
 ================================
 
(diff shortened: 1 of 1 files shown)


apmellot:
More efforts are needed to make the "Visualizing cross-validation behavior in skada" easy to understand.
- What is the interest of doing a DA cross-validation compared to using make_da_pipeline with scikit-learn cross-validation (like in "Using cross_val_score with skada" example)?
- The LeaveOneDomainOut needs more explanations. It is not clear that domains are represented twice (for when they are considered source or target).
- Are the test samples from source domains ever useful? From what I understand the adapter uses source and target train samples, then the base estimator is fitted with the source train samples and finally predict/score is performed on target test samples.

PR #184, opened 2024-04-30
[WIP] Major update of Adapter API

kachayev:
The original idea was to get ability to propagate labels, though it turned out there are a few other things we need to take care about. The list of changes:

* `make_da_pipeline` prepends a transformer to convert input into `MetadataContainer` to carry information about features, labels and additional named parameters (metadata)
* each Selector is responsible for unpacking it before passing arguments to the underlying estimator(s) and merging the result(s) back into the container
*  `adapt` is not longer a part of `BaseAdapter` interface, use `fit_transform` (adaptation during fit-time) and `transform` (predict-time) instead. By default, `BaseAdapter` just pass samples through (without changing them)
* instead of `AdaptationOutput` adapter is suppose to return either a) samples `X`, b) pair `X, params` with dictionary of additional keys (e.g. `sample_weight`), c) triple `X, y, params`

There are a few things I need to take care about:

- [ ] Make `PerDomain` work
- [ ] Make `SelectSource` work
- [ ] Make `SelectTarget` work
- [ ] Make `SelectSourceTarget` work
- [ ] Fix tests for MMD reweight adapter
- [ ] Update all adapters to avoid `adapt`
- [ ] Remove `AdaptationOutput` (maybe a separate task)
- [ ] Properly update documentation regarding the change in the API
- [ ] Properly deal with metadata routing parameter renaming (maybe a separate task)


skada/base.py
```diff
@@ -329,53 +251,36 @@ def decision_function(self, X, **params):
...
-                                f"The adapter provided '{k}' parameter which is not explicitly set as "  # noqa
-                                f"requested or not for '{routing_request.owner}.{method}'.\n"  # noqa
-                                f"Make sure that metadata routing is properly setup, e.g. by calling 'set_{method}_request()'. "  # noqa
-                                "See documentation at https://scikit-learn.org/stable/metadata_routing.html"  # noqa
-                            ) from e
+                # xxx(okachaiev): there's still a way for this to fail,
+                # if non-final estimator consumed the input that was
+                # generated by one of the adapters
+                if isinstance(metadata_container, dict):
+                    iter = metadata_container.items()
```
rflamary, line 321: this was not covered by test and do not looked like an edge

skada/base.py
```diff
@@ -653,17 +584,19 @@ def get_estimator(self, domain: Literal['source', 'target']) -> BaseEstimator:
         check_is_fitted(self)
         return self.estimators_[domain]
 
-    def fit(self, X_container, y, **params):
-        X, params = self._unwrap_from_container(X_container, params)
-        # xxx(okachaiev): seems like we have this block of code rather often
-        #                 maybe should be a helper
+    def fit(self, X_container, y=None, **params):
```
rflamary: fit is not tested here?

skada/base.py
```diff
@@ -677,89 +610,58 @@ def fit(self, X_container, y, **params):
...
-            allow_source=True,
-            **params
-        )
+    @available_if(_estimator_has('transform', base_attr_name='source_estimator'))
+    def fit_transform(self, X_container, y=None, **params):
+        if hasattr(self.source_estimator, "fit_transform"):
+            domain_outputs = self._fit('fit_transform', X_container, y=y, **params)
+            output = _merge_domain_outputs(len(X_container), domain_outputs, allow_containers=True)
+        else:
+            self.fit(X_container, y, **params)
```
rflamary, line 693: not etsted either

skada/tests/test_mapping.py
```diff
@@ -171,23 +192,33 @@ def _base_test_new_X_adapt(estimator, da_dataset):
...
-                marks=pytest.mark.skipif(not torch, reason="PyTorch not installed"),
-            )
-        ),
-        (
-            pytest.param(
-                MMDLSConSMappingAdapter(gamma=1e-3),
-                marks=pytest.mark.skipif(not torch, reason="PyTorch not installed"),
-            )
-        ),
+        # (
```
rflamary, line 195: commented out for debugging? should be brought back no?

skada/tests/test_mapping.py
```diff
@@ -171,23 +192,33 @@ def _base_test_new_X_adapt(estimator, da_dataset):
         (LinearOTMappingAdapter()),
         (CORALAdapter()),
```
rflamary, line 193: We do not need CORALAdapter twice I guess?

skada/tests/test_mapping.py
```diff
@@ -197,11 +228,21 @@ def test_new_X_adapt(estimator, da_reg_datasets):
         EntropicOTMappingAdapter(),
         LinearOTMappingAdapter(),
         CORALAdapter(),
-        pytest.param(
-            MMDLSConSMappingAdapter(gamma=1e-3),
-            marks=pytest.mark.skipif(not torch, reason="PyTorch not installed"),
-        ),
+        # pytest.param(
```
rflamary, line 231: still commented

skada/tests/test_reweight.py
```diff
@@ -207,7 +219,7 @@ def test_new_X_adapt(estimator, da_reg_datasets):
         DiscriminatorReweightAdapter(),
         KLIEPReweightAdapter(gamma=[0.1, 1, "auto", "scale"], random_state=42),
         KMMReweightAdapter(gamma=0.1, smooth_weights=True),
-        MMDTarSReweightAdapter(gamma=1.0),
+        # MMDTarSReweightAdapter(gamma=1.0),
```
rflamary, line 222: shoudl be decommented

rflamary (commented):
Thanks so much @kachayev I have a few comments

skada/tests/test_mapping.py
kachayev: good catch, done

skada/tests/test_mapping.py
kachayev: good catch, done

skada/tests/test_reweight.py
kachayev: oh, yeah. that's the one that fails only when all tests are running but works if you run a single module

skada/tests/test_mapping.py
kachayev: I don't know what was the original idea for the test but all of them are used twice in here. I can clean it up

kachayev: removed duplicates

skada/tests/test_reweight.py
kachayev: this seems to be a problem with dataset generator. if i add

```
    assert not np.any(np.isnan(y_train))
```

to check that we don't get `NaN` labels, many of those tests fail. i'm gonna add separate issue for this. seems completely irrelevant for this PR. wdyt?

rflamary: OK seems irrelevant indeed

rflamary: but we must open a issus to add thsoe etst back

PR #185, opened 2024-04-30
[MRG] Add Amazon review dataset to skada

YanisLalou:
- Same structure as office31


rflamary (approved):

kachayev:
A few comments from #196 also seem to be applicable here.

kachayev (approved):

PR #196, opened 2024-05-03
[MRG] Add officehome dataset

YanisLalou:
Done:
- Fetch zip from fig share + unzip it
- Parse CSV
- Change OfficeHomeDomain values
- Create Test 
- FInd a way to not parse the csv for each new domain (too time intensive)

Done with @ambroiseodt


YanisLalou:
Merge 1st PR #197 to make the tests work

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+) -> Union[Bunch, Tuple[np.ndarray, np.ndarray]]:
+    if not isinstance(domain, OfficeHomeDomain):
+        domain = OfficeHomeDomain(domain.lower())
+    data_home = get_data_home(data_home)
+    dataset_dir = os.path.join(data_home, loader_spec.dataset_dir)
+    if not os.path.exists(dataset_dir):
+        if not download_if_missing:
+            raise OSError("Data not found and `download_if_missing` is False")
+        os.makedirs(dataset_dir)
+        # juggling with `extract_root` is only required because SURF features
```
kachayev, line 229: Based on this comment, seems like `extract_root` key is irrelevant for this dataset, right? 🤔

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+        )
+
+    (X, y) = _load_office_home(
+        dataset_dir,
+        domain,
+        shuffle=shuffle,
+        random_state=random_state,
+    )
+
+    dataset = DomainAwareDataset()
```
kachayev, line 244: Minor, but we don't really need to create dataset object if `return_X_y=True`

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+def _load_office_home(
+    dataset_dir: Union[os.PathLike, str],
+    domain: OfficeHomeDomain,
+    shuffle: bool = False,
+    random_state: Union[None, int, np.random.RandomState] = None,
+) -> Bunch:
+    fullpath = Path(dataset_dir)
+    csv_path = str(fullpath) + "/" + "OfficeHomeResnet50.csv"
```
kachayev, line 270: Because `fullpath` is `Path` object, you can do

```
csv_path = fullpath / "OfficeHomeResnet50.csv"
```

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+
+    with open(csv_path, newline="") as csvfile:
+        # Create a CSV reader object
+        reader = csv.reader(csvfile, quotechar='"')
+
+        # Skip the header row
+        next(reader)
+
+        # Iterate over each row in the CSV file
+        for row in reader:
```
kachayev, line 315: If each row has to have exactly 3 things in it, it might be more readable to say

```
for (feature, label, domain) in reader:
    pass
```

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+            # Append the feature (first column) to the features list
+            features.append(eval(row[0]))
```
kachayev, line 317: Oh, performing `eval` on the downloaded file is a huge security risk. Why do we need it here?

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+
+    X = np.array(features)
+    y = np.array(labels)
+    domains = np.array(domains)
+
+    return (X, y, domains)
+
+
+def shuffle_arrays(*arrays, random_state=None):
+    if random_state is None:
```
kachayev, line 331: We can use `random_state = check_random_state(random_state)` from sklearn to be consistent.

skada/datasets/_office_home.py
```diff
@@ -0,0 +1,338 @@
...
+            domains.append(row[2])
+
+    X = np.array(features)
+    y = np.array(labels)
+    domains = np.array(domains)
+
+    return (X, y, domains)
+
+
+def shuffle_arrays(*arrays, random_state=None):
```
kachayev, line 330: Is this function supposed to be used outside of the module? if no, better to name it with `_shuffle_array`

kachayev (changes requested):

skada/datasets/_office_home.py
YanisLalou: First of all thanks for all the comments 🙏, its always very helpful. But yes basically when I generated the csv file (by doing pd.to_csv()), my the features in my csv were always `str`. So I figured out I could just do an eval, but ok let me take another look at how I generate the csv.

kachayev: If you use `to_csv()` on the Pandas data frame, maybe `read_csv()` would work to reconstruct it back? 🤔

YanisLalou: I agree but pandas is not in the requirements.txt of skada and as far as I know we don't want it. Basically what I'm planning to do is to use a csv with 2048 + 2 columns for features + label/domain, that way I think I won't have to use eval()

kachayev: Ah, your are right. We removed pandas from dependencies. Completely forgot about this.

kachayev (approved):

PR #205, opened 2024-05-23
Fix density reweight

YanisLalou:
Now when `source_weights.mean() == 0` we don't divide `source_weights` by it.

### DIFF
--- a/skada/metrics.py
+++ b/skada/metrics.py
@@ -351,10 +351,7 @@ class DeepEmbeddedValidation(_BaseDomainAwareScorer):
 
     def _no_reduc_log_loss(self, y, y_pred):
         return np.array(
-            [
-                self.cross_entropy_loss(y[i : i + 1], y_pred[i : i + 1])
-                for i in range(len(y))
-            ]
+            [self.cross_entropy_loss(y[i], y_pred[i]) for i in range(len(y))]
         )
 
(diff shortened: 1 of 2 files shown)


skada/metrics.py
```diff
@@ -171,7 +171,10 @@ def _score(self, estimator, X, y, sample_domain=None, **params):
         ws = self.weight_estimator_source_.score_samples(X_source)
         wt = self.weight_estimator_target_.score_samples(X_source)
         weights = np.exp(wt - ws)
-        weights /= weights.sum()
+
+        if weights.sum() != 0:
```
kachayev, line 175: How can we get 0 if weights are exp? 🤔

YanisLalou: I think its due to python precision, I've checked the value manually for an exp and I had at some point `exp(-20_000) = 0`

rflamary: so you should probably use logsumexp that is numrically stable there (susing a concatenation of ws wt)

kachayev: So, if we got all weights 0., the following estimator should fail anyways. We should log warning here, that it's the reweighter who's to blame for the failure.

YanisLalou: I tried with logsumexp but we have the same issue. I propose logging a warning as @kachayev suggested + using uniform weights instead

Issue #177, opened 2024-04-30
Deep method to support non-vectorized feature vectors


kachayev:
Right now we assume that features (e.g. `features_s`, `features_t`) are 2-dimentional tensors. This might not always be the case. Should be either stated in the documentation clearly, or we can make it work for ndim > 2.

PR #218, opened 2024-07-11
[WIP] Add multi layer adaptation possibility + add flattening of each feature outputs

tgnassou:
This PR allows the user to put several layer names to adapt to different network levels, as done in several papers.

This PR also forces the features of each layer to be 2D, answering Issue #177.

It adds a lot for loops in the code; I don't see another way to do it; let me know what you think.


antoinecollas:
@tgnassou Why did you close this pr? Has it already been solved?

PR #219, opened 2024-07-17
[WIP] faster subspace methods (subsampling and other optional parameters)

rflamary:
Subspace methods can have cubic complexity with the number of samples and we believe that subsampling can be an efficient way to learn an approximate subspace


antoinecollas:
we should check https://scikit-learn.org/stable/modules/generated/sklearn.kernel_approximation.Nystroem.html#sklearn.kernel_approximation.Nystroem

rflamary:
Closing this because other more general implementation with Adapter/Transformer is coming

PR #220, opened 2024-07-19
[MRG] Fix batch issue when generating features + add sample_weight in deep models

YanisLalou:
Before when doing a model.predict_features(X) we were passing all the input in a single batch.
This created CUDA out of memory issues when working with big datasets.
Thus here I tried to mimic the behaviour of skorch models to use batching. 

Might not be the best way to fix that issue though, would love your opinion @tgnassou


skada/deep/base.py
```diff
@@ -497,7 +503,41 @@ def predict_features(self, X):
         if not torch.is_tensor(X):
             X = torch.tensor(X)
 
-        _, features = self.module_(
-            X, sample_domain=None, is_fit=False, return_features=True
-        )
-        return features
+
+        features_list = []
+        for features in self.feature_iter(X, training=False, return_features=True):
```
tgnassou, line 508: Don't need to repeat defautl parameters

skada/deep/base.py
```diff
@@ -497,7 +503,41 @@ def predict_features(self, X):
...
+            features = features[0] if isinstance(features, tuple) else features
```
tgnassou, line 509: Why do you test if it is a tuple ? Right now It should return only the features, maybe I'm wrong. I'm working on a PR when now the features could also be a list of features. Let's see how it evolves

tgnassou (commented):
I'm wondering if you can not just create a dataloader on your own, without using skorch function. Like that you only create a dataloader with X, iterate on the dataloader and that's it. It seems very long to do something simple, but maybe I'm wrong

skada/deep/base.py
```diff
@@ -5,38 +5,60 @@
...
         The initialized criterion (loss) used to compute the
         loss to reduce the divergence between domains.
     reg: float, default=1
         Regularization parameter.
+    reduce: bool or None, default=None
+        Whether to reduce the loss. If None, uses the default behavior of the criteria.
     """
 
-    def __init__(self, base_criterion, adapt_criterion, reg=1):
+    def __init__(self, base_criterion, adapt_criterion, reg=1, reduce=False):
```
antoinecollas, line 44: Why do we now need to specify the reduction?

skada/deep/base.py
```diff
@@ -254,18 +276,31 @@ def __sklearn_clone__(self) -> torch.nn.Module:
...
-            features_s = self.intermediate_layers[self.layer_name]
-            y_pred_t = self.base_module_(X_t)
-            features_t = self.intermediate_layers[self.layer_name]
+
+            # Pass sample_weight to base_module_
+            if sample_weight is not None:
+                sample_weight_s = sample_weight[source_idx]
+                sample_weight_t = sample_weight[~source_idx]
+
+                y_pred_s = self.base_module_(X_s, sample_weight=sample_weight_s)
```
antoinecollas, line 293: Why do we need `sample_weight` here? We do not compute loss, right?

skada/deep/base.py
```diff
@@ -301,203 +338,375 @@ def forward(self, X, sample_domain=None, is_fit=False, return_features=False):
             )
         else:
             if return_features:
-                return self.base_module_(X), self.intermediate_layers[self.layer_name]
+                return self.base_module_(X, sample_weight=sample_weight), self.intermediate_layers[self.layer_name]
```
antoinecollas, line 341: Same thing here

skada/deep/base.py
```diff
@@ -254,18 +276,31 @@ def __sklearn_clone__(self) -> torch.nn.Module:
...
+
+            # Pass sample_weight to base_module_
+            if sample_weight is not None:
+                sample_weight_s = sample_weight[source_idx]
+                sample_weight_t = sample_weight[~source_idx]
+
+                y_pred_s = self.base_module_(X_s, sample_weight=sample_weight_s)
+                features_s = self.intermediate_layers[self.layer_name]
+
+                y_pred_t = self.base_module_(X_t, sample_weight=sample_weight_t)
```
antoinecollas, line 296: Same thing here

skada/deep/base.py
```diff
@@ -301,203 +338,375 @@ def forward(self, X, sample_domain=None, is_fit=False, return_features=False):
...
-        #     iterator_valid = None
-        # else:
-        #     iterator_valid = (
-        #         DomainBalancedDataLoader if iterator_valid is None else iterator_valid
-        #     )
-        iterator_train = (
-            DomainBalancedDataLoader if iterator_train is None else iterator_train
-        )
-
+        iterator_train = DomainBalancedDataLoader if iterator_train is None else iterator_train
```
antoinecollas, line 368: Can you put back old code? (to remove unnecessary modifications)

skada/deep/base.py
```diff
@@ -301,203 +338,375 @@ def forward(self, X, sample_domain=None, is_fit=False, return_features=False):
...
-            X = {"X": X}
-            X["sample_domain"] = sample_domain
-
+        # TODO val is not working
+        # if train_split is None:
+        # iterator_valid = None
+        # else:
+        # iterator_valid = (
+        # DomainBalancedDataLoader if iterator_valid is None else iterator_valid
+        # )
```
antoinecollas, line 402: Why did this comment move?

antoinecollas (commented):
`sample_weight` is everywhere but in practice we only use it to reweight the loss, right? IMO, it should used only when calling the loss but maybe I misunderstood something.

If you run coverage run -m `pytest -v -s && coverage html && open htmlcov/index.html` on your branch and on the main branch, you will see that you have added 24 lines in `skada/deep/base.py` that are not covered by tests.

antoinecollas:
From [Skorch FAQ](https://skorch.readthedocs.io/en/stable/user/FAQ.html#i-want-to-use-sample-weight-how-can-i-do-this): when X is a dict, its keys are passed as kwargs to `forward`, thus our forward has to have the arguments 'data' and 'sample_weight'; usually, sample_weight can be ignored here.

tgnassou:
Very good PR, the only thing missing is to modify also `CDANModule`

PR #221, opened 2024-08-11
[MERGE] Add new scorer: MixValScorer

YanisLalou:
Implementation of the paper:
https://proceedings.neurips.cc/paper_files/paper/2023/file/7721f1fea280e9ffae528dc78c732576-Paper-Conference.pdf

### DIFF
--- a/skada/metrics.py
+++ b/skada/metrics.py
@@ -12,7 +12,7 @@ from copy import deepcopy
 import numpy as np
 from sklearn.base import BaseEstimator, clone
 from sklearn.linear_model import LogisticRegression
-from sklearn.metrics import accuracy_score, balanced_accuracy_score, check_scoring
+from sklearn.metrics import balanced_accuracy_score, check_scoring
 from sklearn.model_selection import train_test_split
 from sklearn.neighbors import KernelDensity
 from sklearn.preprocessing import LabelEncoder, Normalizer
@@ -632,7 +632,7 @@ class MixValScorer(_BaseDomainAwareScorer):
 
     Parameters
     ----------
-    lmbd : float, default=0.55
+    alpha : float, default=0.55
         Mixing parameter for mixup.
     random_state : int, RandomState instance or None, default=None
         Controls the randomness of the mixing process.
@@ -641,7 +641,7 @@ class MixValScorer(_BaseDomainAwareScorer):
 
     Attributes
     ----------
-    lmbd : float
+    alpha : float
         Mixing parameter.
     random_state : RandomState
         Random number generator.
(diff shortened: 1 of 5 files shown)


skada/metrics.py
```diff
@@ -619,3 +619,125 @@ def _score(self, estimator, X, y, sample_domain=None):
...
+    MixVal scorer for unsupervised domain adaptation.
+
+    This scorer uses mixup to create mixed samples from the target domain,
+    and evaluates the model's consistency on these mixed samples.
+
+    See [32]_ for details.
+
+    Parameters
+    ----------
+    lmbd : float, default=0.55
```
tgnassou, line 635: I don't know if sklearn uses `lmbd` name. Maybe switch to alpha or mixup_param is better ?

skada/metrics.py
```diff
@@ -619,3 +619,125 @@ def _score(self, estimator, X, y, sample_domain=None):
...
+        # Generate mixed samples
+        X_mixed, y_mixed = self._generate_mixed_samples(X[~source_idx], y_pred)
+
+        # Get predictions for mixed samples
+        y_pred_mixed = estimator.predict_proba(
+            X_mixed, sample_domain=np.full(X_mixed.shape[0], -1)
+        )
+
+        # Calculate ICE score
+        ice_score = accuracy_score(y_mixed.argmax(axis=1), y_pred_mixed.argmax(axis=1))
```
tgnassou, line 737: In the paper, they don't compute accuracy on all the mixed samples. They separate samples with the same `y_pred` and a different one. Then, they compute the accuracy score for diff_mixed_samples and same_mixed_sample and take the mean of the two scores. See Appendix A of the paper.

skada/metrics.py
```diff
@@ -619,3 +619,131 @@ def _score(self, estimator, X, y, sample_domain=None):
...
+        X_target = X[~source_idx]
+
+        rng = check_random_state(self.random_state)
+        rand_idx = rng.permutation(X_target.shape[0])
+
+        # Get predictions for target samples
+        pred_a = estimator.predict_proba(
+            X_target, sample_domain=sample_domain[~source_idx]
+        )
+        pl_a = pred_a.argmax(axis=1)
```
tgnassou, line 712: I don't really like the name `pl_a` and `pl_b` to be consistent we can call them `labels_a` and `labels_b`?

skada/metrics.py
```diff
@@ -619,3 +619,131 @@ def _score(self, estimator, X, y, sample_domain=None):
...
+        pl_b = pl_a[rand_idx]
+
+        # Intra-cluster and inter-cluster mixup
+        same_idx = (pl_a == pl_b).nonzero()[0]
+        diff_idx = (pl_a != pl_b).nonzero()[0]
+
+        # Mixup with images and hard pseudo labels
+        mix_inputs = self.alpha * X_target + (1 - self.alpha) * X_target[rand_idx]
+        mix_labels = pl_a if self.alpha > 0.5 else pl_b
```
tgnassou, line 721: In the pseudo-code they are doing that, but neither in the paper nor in the GitHub they are using this formula for the mix_labels. They're using the formula with `alpha`, maybe it is the same ?

YanisLalou: Yes I agree, in the paper they're using the same formula as for X. However before going forward with the code, we need to agree what we should follow for all the implementation: The paper, the pseudo-code or the git code ?

PR #222, opened 2024-08-15
[WIP] Fix mixval

antoinecollas:


skada/metrics.py
```diff
@@ -713,24 +722,30 @@ def _score(self, estimator, X, y=None, sample_domain=None, **params):
...
 
-        # Mixup with images and hard pseudo labels
+        # Mixup with X_target and hard pseudo labels
         mix_inputs = self.alpha * X_target + (1 - self.alpha) * X_target[rand_idx]
-        mix_labels = self.alpha * labels_a + (1 - self.alpha) * labels_b
-
-        # Obtain predictions for the mixed samples
-        mix_pred = estimator.predict(
-            mix_inputs, sample_domain=np.full(mix_inputs.shape[0], -1)
+        lb = LabelBinarizer(neg_label=False, pos_label=True).fit(y[source_idx])
```
YanisLalou, line 732: Not sure that LabelBinarizer will work everytime, why don't you use OneHotEncoder?

antoinecollas: One hot encoder is made for X not y. But indeed we should change the test for multiple classes.

skada/metrics.py
```diff
@@ -713,24 +727,30 @@ def _score(self, estimator, X, y=None, sample_domain=None, **params):
...
+        one_hot_labels_a = lb.transform(labels_a)
+        one_hot_labels_b = lb.transform(labels_b)
+        mix_labels = np.argmax(
+            self.alpha * one_hot_labels_a + (1 - self.alpha) * one_hot_labels_b, axis=1
         )
```
YanisLalou, line 737: Oh I just saw that you were doing a `np.argmax()` here. It will work yes, but I feel like it'll be much easier/faster to just implement directly the if alpha >= 0.5 condition instead of computing the onehotencoding and then going back to the categorical labels

PR #235, opened 2024-09-13
[MRG] Fix order of feature acquisition for deep module

tgnassou:


antoinecollas:
is it possible to add a unit test?

antoinecollas (approved):
Good catch @tgnassou ! I just added a test. It doesn't pass without your commit but passes with!

PR #238, opened 2024-09-19
[WIP] Add pretrain param for deep method

tgnassou:


skada/deep/base.py
```diff
@@ -563,6 +566,24 @@ def fit(
         # but it requires to adapt skada.utils.check_X_y_domain
         # to handle dict, Dataset, torch.Tensor, ...
 
+        if self.epochs_pretrain_on_source:
+            # Pretrain on source domain
+
+            net = NeuralNetClassifier(
```
YanisLalou, line 572: Should also put the `net` on the same device as `self`.

PR #246, opened 2024-09-25
[TO_REVIEW] Fix deep coral

antoinecollas:


skada/deep/losses.py
```diff
@@ -40,10 +41,11 @@ def deepcoral_loss(features, features_target, assume_centered=False):
     if not assume_centered:
         features = features - features.mean(0)
         features_target = features_target - features_target.mean(0)
-    cov = torch.cov(features)
-    cov_target = torch.cov(features_target)
-    diff = cov - cov_target
-    loss = (diff * diff).sum() / (4 * len(cov) ** 2)
+    cov = torch.cov(features.T)
+    cov_target = torch.cov(features_target.T)
+    divergence = torch.norm(cov - cov_target, p="fro") ** 2
```
agramfort, line 46:
```suggestion
    from torch.nn.functional import mse_loss
    divergence = mse_loss(cov, cov_target)
```
should be slightly more efficient

skada/deep/losses.py
```diff
@@ -43,7 +44,7 @@ def deepcoral_loss(features, features_target, assume_centered=False):
         features_target = features_target - features_target.mean(0)
     cov = torch.cov(features.T)
     cov_target = torch.cov(features_target.T)
-    divergence = torch.norm(cov - cov_target, p="fro") ** 2
+    divergence = mse_loss(cov, cov_target)
```
rflamary, line 47: won't mse lead to scaling problems?

PR #247, opened 2024-09-30
[MRG] fix `predict_proba` for deep method

tgnassou:
The `predict_proba` of deep method was choosing the nonlin attributes by looking at `criterion` attribute of the solvers. The `DomainAwareNet` has a `DomainAwareCriterion` as criterion. We should choose the nonlin base on `criterion.base_criterion`. This PR does that.

### DIFF
--- a/skada/deep/base.py
+++ b/skada/deep/base.py
@@ -7,13 +7,14 @@
 
 from abc import abstractmethod
 from typing import Dict, Any, Union
+from functools import partial
 
 import torch
 from torch.utils.data import DataLoader, Sampler, Dataset
 from sklearn.base import _clone_parametrized
 from skorch import NeuralNetClassifier
 
-from .utils import _register_forwards_hook
+from .utils import _register_forwards_hook, _infer_predict_nonlinearity
 
 from skada.base import _DAMetadataRequesterMixin
 
@@ -621,6 +622,7 @@ class DomainAwareNet(NeuralNetClassifier, _DAMetadataRequesterMixin):
             The predicted class probabilities.
         """
         X, _ = self._prepare_input(X, sample_domain, sample_weight)
+        self.predict_nonlinearity = _infer_predict_nonlinearity(self)
         return super().predict_proba(X, **predict_params)
 
     def predict_features(self, X: Union[Dict, torch.Tensor, np.ndarray, Dataset]):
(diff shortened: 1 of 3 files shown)


skada/deep/utils.py
```diff
@@ -52,3 +55,33 @@ def check_generator(seed):
...
+    criterion = getattr(net, net._criteria[0] + '_').base_criterion
+
+    if isinstance(criterion, CrossEntropyLoss):
+        return partial(torch.softmax, dim=-1)
+
+    if isinstance(criterion, BCEWithLogitsLoss):
+        return _sigmoid_then_2d
+
+    if isinstance(criterion, BCELoss):
+        return _make_2d_probs
```
YanisLalou, line 85: Here shouldnt be inverted? The `BCEWithLogitsLoss` "combines a Sigmoid layer and the BCELoss in one single class".
Thus having `_sigmoid_then_2d` for `BCELoss` and `_make_2d_probs` for `BCEWithLogitsLoss`.

tgnassou: I copied/pasted the skorch function

tgnassou: But yeah it's true. I think they made a mistake

antoinecollas:
Right now, it's not possible to do classification with only two classes in `skada.deep`, see #249 .
Thus, we should only check is `criterion` is `CrossEntropyLoss` (and not `BCELoss` or `BCEWithLogitsLoss`).

antoinecollas:
I added a test that checks when `criterion` is `CrossEntropyLoss`. I have no idea how to test the other cases...

agramfort (commented):
looks a bit too magical for me...

skada/deep/base.py
```diff
@@ -621,6 +622,7 @@ def predict_proba(
             The predicted class probabilities.
         """
         X, _ = self._prepare_input(X, sample_domain, sample_weight)
+        self.predict_nonlinearity = _infer_predict_nonlinearity(self)
```
agramfort, line 625: I don't like that the instance is modified in the predict function. One poor man solution would be to delete the attribute after the super().predict_proba call

antoinecollas: I removed the `self.predict_nonlinearity =...`

PR #250, opened 2024-10-07
[MRG] Add MCC method

tgnassou:


skada/deep/losses.py
```diff
@@ -209,3 +209,63 @@ def forward(
...
+    .. [33] Ying Jin, Ximei Wang, Mingsheng Long, Jianmin Wang.
+            Minimum Class Confusion for Versatile Domain Adaptation.
+            In ECCV, 2020.
+    """
+    # Probability Rescaling
+    y_scaled = probability_scaling(y, temperature=T)
+
+    # Uncertainty Reweighting & class correlation matrix
+    H = -torch.sum(y_scaled * torch.log(y_scaled), axis=1)
+    W = (1 + torch.exp(H)) / torch.mean(1 + torch.exp(H))
```
YanisLalou, line 260:
```suggestion
    W = (1 + torch.exp(-H)) / torch.mean(1 + torch.exp(-H))
```

skada/deep/losses.py
```diff
@@ -209,3 +209,63 @@ def forward(
...
+    y_weighted = torch.matmul(torch.diag(W), y_scaled)
+    C = torch.einsum("ij,ik->jk", y_scaled, y_weighted)
+
+    # Category Normalization
+    C_tilde = C / torch.sum(C, axis=1, keepdim=True)
+
+    # MCC Loss
+    C_ = C_tilde - torch.diag(torch.diag(C_tilde))
```
YanisLalou, line 268:
```suggestion
    C_ = C_tilde - torch.diag(C_tilde)
```

YanisLalou (commented):
2 small suggested changes - otherwise looks good to me

skada/deep/losses.py
tgnassou: Good  catch

skada/deep/losses.py
tgnassou: What I want is to remove the diag. If I do only one torch.diag I will have a vector and then remove the diag on all the line. Two torch.diag create a diagonal matrix

YanisLalou: Ah my bad I thought it returned a diagonal matrix directly

PR #256, opened 2024-10-15
[TO_REVIEW] Bug fix when None in make_da_pipeline

antoinecollas:
**Enhancement: Add compatibility with `None` in `skada` pipelines**

This PR introduces support for using `None` within `skada` pipelines, similar to how `scikit-learn` handles it. For example:

```Python
pipe = make_pipeline(
        StandardScaler(),
        None,  # Skip step
        PCA(),
        LogisticRegression(),
)
```

Now, `skada` pipelines can also include `None` as a step:

```Python
pipe = make_da_pipeline(
        StandardScaler(),
        None,  # Skip step
        PCA(),
        SubspaceAlignmentAdapter(n_components=2),
        LogisticRegression(),
)
```

### DIFF
--- a/skada/_pipeline.py
+++ b/skada/_pipeline.py
@@ -108,7 +108,7 @@ def _wrap_with_selector(
     estimator: BaseEstimator,
     selector: Union[str, Callable[[BaseEstimator], BaseSelector]],
 ) -> BaseSelector:
-    if not isinstance(estimator, BaseSelector):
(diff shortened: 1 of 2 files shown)


rflamary (approved):
LGTM, I would maybe test with and without the None to check that we have the same output.

agramfort (approved):

PR #259, opened 2024-10-23
[MRG] Subsampling transformer

rflamary:
The main idea is to provide a Subsampler that can be done in a pipeline in order to speedup some methods that do not scale well with number of samples.

### DIFF
--- a/skada/deep/base.py
+++ b/skada/deep/base.py
@@ -11,11 +11,13 @@ from typing import Dict, Any, Union
 import torch
 from torch.utils.data import DataLoader, Sampler, Dataset
 from sklearn.base import _clone_parametrized
+from sklearn.metrics import accuracy_score
 from skorch import NeuralNetClassifier
 
 from .utils import _register_forwards_hook
 
 from skada.base import _DAMetadataRequesterMixin
+from skada.utils import check_X_domain
 
 import numpy as np
 from skorch.utils import to_device
@@ -495,7 +497,7 @@ class DomainAwareModule(torch.nn.Module):
 
 class DomainAwareNet(NeuralNetClassifier, _DAMetadataRequesterMixin):
     __metadata_request__fit = {"sample_weight": True}
-    __metadata_request__score = {"sample_weight": True}
+    __metadata_request__score = {'sample_weight': True, 'sample_domain': True, 'allow_source': True}
     """
     A domain-aware neural network classifier with sample weight support.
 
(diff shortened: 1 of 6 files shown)


skada/transformers.py
```diff
@@ -0,0 +1,66 @@
...
+        self.random_state = random_state
+
+    def _pack_params(self, idx, **params):
+        return {
+            k: (v[idx] if idx is not None else v)
+            for k, v in params.items()
+            if v is not None
+        }
+
+    def fit_transform(self, X, y=None, *, sample_domain=None, sample_weight=None):
```
kachayev, line 41: I guess it would be nice to have test case to check if `sample_weight` is processed properly, as right now we only have checks in place for `sample_domain`

skada/transformers.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.n_subsample >= X.shape[0]:
+            return (
+                X,
+                y,
+                self._pack_params(
+                    None, sample_domain=sample_domain, sample_weight=sample_weight
+                ),
+            )
+
+        idx = self.rng_.choice(X.shape[0], self.n_subsample, replace=False)
```
kachayev, line 56: In practical scenarios, do we need the sampling to be done per-domain? 🤔 It might be the case that the result of sampler omits domains with few(-er) entries.

rflamary: In this case do we create a DomainStratifiedSampleTransformer or do we add a parameter?  In any Case we should probably use sklearn or skada Splitters here

kachayev: I'm always for having separate classes if the use case is clear. Much easier to just open the list of things to spot what you're looking for! `DomainStratifiedSampleTransformer` sounds great.

skada/tests/test_transformers.py
```diff
@@ -0,0 +1,98 @@
...
+
+    transformer = DomainStratifiedSubsampleTransformer(train_size=train_size)
+
+    X_subsampled, y_subsampled, params = transformer.fit_transform(
+        X, y, sample_domain=sample_domain
+    )
+
+    assert X_subsampled.shape[0] == train_size
+    assert y_subsampled.shape[0] == train_size
+    assert "sample_domain" in params
```
antoinecollas, line 71: check shape of params["sample_domain"]

antoinecollas: also, we should check that `sample_domain` is stratified

Issue #16, opened 2023-11-21
`check_*_domain` should fail when the same domain is given as a source and a target


kachayev:
This check is currently not implemented.

Rouxben:
hey, sorry but I'm not sure I understood what the function is exactly supposed to do:
is it supposed to check if two domains have the same name (1 and -1 as an exemple),
or is it supposed to check if two domains, one from target and another from source, have the same elements

kachayev:
@BuenoRuben The functions are already present in the code, so you can review them to understand their purposes. Given that they contain a significant number of internal clauses, it might be more straightforward to comprehend their functionality by reading the code directly.

* `check_X_y_domain` [[link](https://github.com/scikit-adaptation/skada/blob/008306a918d5bd48ac670db1716b5c01dd024364/skada/_utils.py#L40C5-L40C21)]
* `check_X_domain` [[link](https://github.com/scikit-adaptation/skada/blob/008306a918d5bd48ac670db1716b5c01dd024364/skada/_utils.py#L105)]

The overall purpose of these functions is similar to the `check_array` function in `sklearn`. Depending on the provided flags, they can execute a wide range of checks. In our case, it’s crucial to ensure that all the checks are: a) necessary, meaning they make sense and are utilized in the code; b) implemented correctly, which we can verify by creating test cases for each check; and c) properly documented in the docstrings, with `sklearn` serving as an excellent reference for how to structure these. Additionally, we need to relocate these functions to a publicly accessible `utils.py` module, as they are intended for both internal use and external use (by anyone developing their own estimators).

kachayev:
@BuenoRuben The idea here is that a specific domain should be included either as source or as a target. Which means that the domain id is either consistently positive or consistently negative. Let's say the id of the domain is 3. It should either be 3 (source) or -3 (target). There shouldn't be a situation where we have both 3 and -3 in the `sample_domain`.

kachayev:
The one place where this happens (and is necessary) is Leave-One-Domain-Out splitter for model selection.

BuenoRuben:
Thanks you for your answer, I will take a look at those functions

PR #261, opened 2024-10-24
[MRG] Check if sample_domain have only unique domains indexes in check_*_domain

apmellot:
In response to #16 

@kachayev is that what you had in mind ?
Should `check_X_y_domain` and `check_X_domain` stay two separate functions?

A test should be added for this situation.


kachayev:
Yes, Apolline, exactly!

Answering your question: `check_X_y_domain` and `check_X_domain` should stay separate as they are designed for different use cases/inputs. Their implementation, ideally, would re-use as much common functionality as possible, for sure, but that's, you know, in an ideal world :)

A few comments:
* i think we need new `allow_` parameter so that users can decide when this should or should not be enforced
* it might be helpful to mention explicitly in the error message that the same domain should not be used both as source and as a target, otherwise it could be confusing

P.S. If I recall correctly, @rflamary had rather strong opinion around the indexing. I'm gonna let him to chime in here. Likely `allow_` should be `True` by default ;)

rflamary (approved):
LGTM especially when keeping the default to True. This will allows us to set this parameter to DAestimators (by keeping the default to TRue) which can allow user to detect problems depending on their choice of numbering for domains in a  future PR.

PR #268, opened 2024-10-25
[MRG] Add `DomainAndLabelStratifiedSubsampleTransformer` + Fix `DomainStratifiedSubsampleTransformer`

YanisLalou:
- Fix `TypeError: DomainStratifiedSubsampleTransformer.transform() got an unexpected keyword argument 'allow_source'`with `DomainStratifiedSubsampleTransformer` 
- Add `DomainAndLabelStratifiedSubsampleTransformer` to have a transformer preserving the percentage of samples
    for each class and for each sample domain (To prevent issues like: `ValueError: This solver needs samples of at least 2 classes in the data, but the data contains only one class: 1`)

### DIFF
--- a/skada/tests/test_transformers.py
+++ b/skada/tests/test_transformers.py
@@ -9,20 +9,20 @@ from sklearn.preprocessing import StandardScaler
 
 from skada import CORAL, make_da_pipeline
 from skada.transformers import (
-    DomainAndLabelStratifiedSubsampleTransformer,
-    DomainStratifiedSubsampleTransformer,
-    SubsampleTransformer,
+    DomainSubsampler,
+    StratifiedDomainSubsampler,
+    Subsampler,
 )
 
 
-def test_SubsampleTransformer(da_dataset):
+def test_Subsampler(da_dataset):
     X, y, sample_domain = da_dataset.pack_train(as_sources=["s"], as_targets=["t"])
     sample_weight = np.ones_like(y)
 
     train_size = 10
 
     # test size of output on fit_transform
-    transformer = SubsampleTransformer(train_size=train_size, random_state=42)
+    transformer = Subsampler(train_size=train_size, random_state=42)
 
     X_subsampled, y_subsampled, params = transformer.fit_transform(
         X, y, sample_domain=sample_domain, sample_weight=sample_weight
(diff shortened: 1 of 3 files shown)


skada/transformers.py
```diff
@@ -129,6 +132,72 @@ def fit_transform(self, X, y=None, *, sample_domain=None, sample_weight=None):
...
+                self._pack_params(
+                    None, sample_domain=sample_domain, sample_weight=sample_weight
+                ),
+            )
+
+        splitter = StratifiedDomainShuffleSplit(
+            n_splits=1, train_size=self.train_size, random_state=self.rng_
+        )
+
+        train_idx, _ = next(splitter.split(X, y, sample_domain))
```
antoinecollas, line 191: `StratifiedDomainShuffleSplit` only statifies w.r.t. the labels but not the domains right? I'm not sure the stratification is done both on domains and labels.

skada/tests/test_transformers.py
```diff
@@ -89,4 +92,50 @@ def test_DomainStratifiedSubsampleTransformer(da_dataset):
...
+    X_subsampled, y_subsampled, params = transformer.fit_transform(
+        X, y, sample_domain=sample_domain, sample_weight=sample_weight
+    )
+
+    assert X_subsampled.shape == (train_size, X.shape[1])
+    assert y_subsampled.shape[0] == train_size
+    assert params["sample_domain"].shape[0] == train_size
+    assert params["sample_weight"].shape[0] == train_size
+    # check stratification
+    assert sum(params["sample_domain"] == 1) == train_size // 2
```
antoinecollas, line 120: you should better check the stratification, i.e. compute frequency per (domain, label)

skada/transformers.py
YanisLalou: `StratifiedDomainShuffleSplit` already stratifies w.r.t labels and domains:
```
This cross-validation object returns stratified randomized folds.
The folds are made by preserving the percentage of samples
for each class and for each sample domain.
```

skada/tests/test_transformers.py
YanisLalou: Done

antoinecollas:
Then, we have to be consistent with the splitters. IMO, we have to rename the subsamplers as:
- DomainStratifiedSubsampleTransformer --> DomainSubsampleTransformer
- DomainAndLabelStratifiedSubsampleTransformer --> StratifiedDomainSubsampleTransformer

antoinecollas:
BTW, we can even get rid of 'Transformer':
- DomainStratifiedSubsampleTransformer --> DomainSubsampler
- DomainAndLabelStratifiedSubsampleTransformer --> StratifiedDomainSubsampler

YanisLalou:
I agree

antoinecollas (approved):

PR #275, opened 2024-11-12
[TO_REVIEW] Remove redundant centroid computation in spherical k-means

YanisLalou:
This pull request includes changes to improve the spherical k-means clustering implementation.

Refactoring of clustering process:

* When using pre-initialised centroids, we do not run the k-means algorithm multiple times. Indeed in that case, the centroids will always have the same position.

Addition of helper method:

* Introduced `_run_single_kmean` method to encapsulate the logic for running a single spherical k-means clustering iteration, which includes computing dissimilarities, updating centroids, and checking for convergence. [[1]](diffhunk://#diff-7424c2462888e7d03099300b412067b00f00c28655a03e22dbc5b94b3d32686fL201-R241) [[2]](diffhunk://#diff-7424c2462888e7d03099300b412067b00f00c28655a03e22dbc5b94b3d32686fL227-R263)

### DIFF
--- a/skada/deep/_divergence.py
+++ b/skada/deep/_divergence.py
@@ -1,6 +1,7 @@
 # Author: Theo Gnassounou <theo.gnassounou@inria.fr>
 #         Remi Flamary <remi.flamary@polytechnique.edu>
 #         Yanis Lalou <yanis.lalou@polytechnique.edu>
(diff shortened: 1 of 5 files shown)


YanisLalou (commented):
LGTM

PR #276, opened 2024-11-14
[MRG] Add SPA method

tgnassou:
https://arxiv.org/pdf/2310.17594

### DIFF
--- a/skada/_mapping.py
+++ b/skada/_mapping.py
@@ -97,9 +97,11 @@ class BaseOTMappingAdapter(BaseAdapter):
         # thus there's no need to perform any transformations
         if X_source.shape[0] > 0:
             X_source = self.ot_transport_.transform(Xs=X_source)
-        X_adapt, _ = source_target_merge(
-            X_source, X_target, sample_domain=sample_domain
-        )
+            X_adapt, _ = source_target_merge(
+                X_source, X_target, sample_domain=sample_domain
+            )
+        else:
+            X_adapt = X_target
         return X_adapt
 
     @abstractmethod
--- a/skada/_mapping.py
+++ b/skada/_mapping.py
@@ -95,13 +95,9 @@ class BaseOTMappingAdapter(BaseAdapter):
         X_source, X_target = source_target_split(X, sample_domain=sample_domain)
         # in case of prediction we would get only target samples here,
         # thus there's no need to perform any transformations
-        if X_source.shape[0] > 0:
-            X_source = self.ot_transport_.transform(Xs=X_source)
-            X_adapt, _ = source_target_merge(
-                X_source, X_target, sample_domain=sample_domain
-            )
-        else:
-            X_adapt = X_target
+        X_adapt, _ = source_target_merge(
+            X_source, X_target, sample_domain=sample_domain
+        )
         return X_adapt
 
     @abstractmethod
--- a/skada/_mapping.py
+++ b/skada/_mapping.py
@@ -95,6 +95,8 @@ class BaseOTMappingAdapter(BaseAdapter):
         X_source, X_target = source_target_split(X, sample_domain=sample_domain)
         # in case of prediction we would get only target samples here,
         # thus there's no need to perform any transformations
+        if X_source.shape[0] > 0:
+            X_source = self.ot_transport_.transform(Xs=X_source)
         X_adapt, _ = source_target_merge(
             X_source, X_target, sample_domain=sample_domain
         )
--- a/skada/deep/_graph_alignment.py
+++ b/skada/deep/_graph_alignment.py
@@ -3,7 +3,6 @@
 # License: BSD 3-Clause
 
 import torch
-from torch import nn
 
 from skada.deep.base import (
     BaseDALoss,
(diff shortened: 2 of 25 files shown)


skada/deep/losses.py
```diff
@@ -414,3 +408,122 @@ def mcc_loss(y, T=1, eps=1e-7):
...
+    memory_outputs : torch.Tensor
+        Memory outputs.
+    K : int, default=5
+        The number of nearest neighbors.
+    sample_idx : torch.Tensor, default=None
+        The sample indices in the batch.
+    """
+    dis = -torch.mm(features_t.detach(), memory_features.t())
+    for i in range(dis.size(0)):
+        dis[i, sample_idx[i]] = torch.inf
```
YanisLalou, line 520: Can't we do instead: `dis.scatter_(1, sample_idx, torch.inf)`

skada/deep/_graph_alignment.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        domain_label_target = torch.ones(
+            (domain_pred_t.size()[0]),
+            device=domain_pred_t.device,
+        )
+
+        # update classification function
+        loss_adv = self.domain_criterion_(
+            domain_pred_s, domain_label
+        ) + self.domain_criterion_(domain_pred_t, domain_label_target)
```
antoinecollas, line 92: in eq (1) of the paper, they use both the cross entropy and the modified cross entropy loss (log(1-prob))

antoinecollas: we should also use label smoothing as the code of the paper: https://github.com/CrownX/SPA/blob/main/code/loss.py#L65-L84

skada/_mapping.py
```diff
@@ -97,9 +97,11 @@ def transform(
...
         if X_source.shape[0] > 0:
             X_source = self.ot_transport_.transform(Xs=X_source)
-        X_adapt, _ = source_target_merge(
-            X_source, X_target, sample_domain=sample_domain
-        )
+            X_adapt, _ = source_target_merge(
+                X_source, X_target, sample_domain=sample_domain
+            )
+        else:
+            X_adapt = X_target
```
antoinecollas, line 104: I think the [#280](https://github.com/scikit-adaptation/skada/pull/280) solved this problem, right?

skada/deep/_graph_alignment.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         reg_adv=1,
         reg_gsa=1,
         reg_nap=1,
+        max_iter=100,
```
antoinecollas, line 56: max_epochs?

skada/deep/_graph_alignment.py
```diff
@@ -87,19 +91,28 @@ def forward(
         )
 
         # update classification function
-        loss_adv = self.domain_criterion_(
-            domain_pred_s, domain_label
-        ) + self.domain_criterion_(domain_pred_t, domain_label_target)
+        loss_adv = (
+            self.reg_adv
+            * eff
```
antoinecollas, line 96: We did not put a scheduler in other adversarial methods. I think you should remove it to be consistent.

skada/deep/_graph_alignment.py
```diff
@@ -66,6 +67,8 @@ def __init__(
         self.K = K
         self.memory_features = memory_features
         self.memory_outputs = memory_outputs
+        self.max_iter = max_iter
+        self.n_iter = 0
```
antoinecollas, line 71: n_epoch?

skada/_mapping.py
```diff
@@ -95,8 +95,6 @@ def transform(
         X_source, X_target = source_target_split(X, sample_domain=sample_domain)
         # in case of prediction we would get only target samples here,
         # thus there's no need to perform any transformations
-        if X_source.shape[0] > 0:
-            X_source = self.ot_transport_.transform(Xs=X_source)
```
antoinecollas, line 99: you want to remove these two lines ?

skada/deep/losses.py
```diff
@@ -414,3 +408,109 @@ def mcc_loss(y, T=1, eps=1e-7):
...
+    memory_outputs : torch.Tensor
+        Memory outputs.
+    sample_idx : torch.Tensor
+        The sample indices in the batch.
+    K : int, default=5
+        The number of nearest neighbors.
+    """
+    dis = -torch.mm(features_t.detach(), memory_features.t())
+
+    dis[torch.arange(dis.size(0)), sample_idx] = float("-inf")
```
antoinecollas, line 503: it should be +inf instead of -inf?

skada/deep/losses.py
```diff
@@ -414,3 +408,109 @@ def mcc_loss(y, T=1, eps=1e-7):
...
+
+    # Get top-K neighbors
+    _, top_k_indices = torch.topk(dis, k=K, dim=1)
```
antoinecollas, line 506: it should be `torch.topk(-dis, k=K, dim=1)`, right?

skada/deep/losses.py
YanisLalou: I agree

skada/deep/callbacks.py
```diff
@@ -73,3 +74,104 @@ def on_epoch_begin(self, net, dataset_train=None, **kwargs):
             target_kmeans.fit(features_t)
 
         net.criterion__adapt_criterion.target_kmeans = target_kmeans
+
+
+class ComputeMemoryBank(Callback):
```
antoinecollas, line 79: Minor remark: I think `ComputeMemoryBank` and `MemoryBankInit` should be merged. Wdyt?

YanisLalou:
Nice results with SPA, lets merge this now!

PR #277, opened 2024-11-15
[MRG] Fix mdd loss

antoinecollas:
Use equations (26)-(29) from https://arxiv.org/pdf/1904.05801

### DIFF
--- a/skada/deep/_adversarial.py
+++ b/skada/deep/_adversarial.py
@@ -3,7 +3,6 @@
 #         Antoine Collas <contact@antoinecollas.fr>
 #
 # License: BSD 3-Clause
-import copy
 import math
 
 import numpy as np
@@ -384,14 +383,6 @@ class MDDLoss(BaseDALoss):
 
     Parameters
     ----------
-    disc_criterion_s : torch criterion (class)
-        The criterion (loss) used to compute the
-        MDD loss for the discriminator on the source domain.
-        If None, a CrossEntropyLoss is used.
-    disc_criterion_t : torch criterion (class)
-        The criterion (loss) used to compute the
-        MDD loss for the discriminator on the target domain.
-        If None, a modified CrossEntropyLoss is used.
     gamma : float (default=4.0)
         Margin parameter following [35]_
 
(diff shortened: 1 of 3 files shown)


antoinedemathelin:
It seems nice, but why adding the new arguments domain_criterion_t and domain_criterion_s? I feel it adds too much complexity. Maybe we can say that the original MDD paper only focus on multi classification, and we do the same here. We can almost remove the domain_criterion argument and hardset it to be crossentropy for source and custom crossentropy for target. 
Wdyt?

We can let to future PRs the generalization of MDD to other losses?

antoinecollas:
I agree, I change this

skada/deep/_adversarial.py
```diff
@@ -421,11 +429,13 @@ def forward(
         pseudo_label_s = torch.argmax(y_pred_s, axis=-1)
         pseudo_label_t = torch.argmax(y_pred_t, axis=-1)
 
-        disc_loss_src = self.disc_criterion_(disc_pred_s, pseudo_label_s)
-        disc_loss_tgt = self.disc_criterion_(disc_pred_t, pseudo_label_t)
+        disc_criterion_s_ = torch.nn.CrossEntropyLoss()
+        disc_loss_s = disc_criterion_s_(disc_pred_s, pseudo_label_s)
+        disc_criterion_t_ = ModifiedCrossEntropyLoss()
+        disc_loss_t = disc_criterion_t_(disc_pred_t, pseudo_label_t)
```
antoinedemathelin, line 435: Is it ok to instantiate functions in the forward pass (I mean from the computational efficiency point of view)?
Maybe it is better to define the criterions in the init as attributes and then use the same in forward?

antoinedemathelin:
Ok, I just have one comment, but the rest is ok for me.

skada/deep/_adversarial.py
```diff
@@ -370,6 +371,22 @@ def CDAN(
...
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, input, target):
+        """Compute the modified CrossEntropyLoss"""
+        prob = F.softmax(input, dim=-1)
+        prob = prob[..., target]
+        log_one_minus_prob = torch.log(1 - prob)
```
tomMoral, line 386: Maybe use this one: https://pytorch.org/docs/stable/generated/torch.log1p.html

PR #279, opened 2024-11-18
[MRG] Apply label smoothing to stabilize MDD

antoinecollas:


skada/deep/_adversarial.py
```diff
@@ -372,18 +372,49 @@ def CDAN(
...
+    """Modified CrossEntropyLoss as described in (29) from [35]_ with label smoothing.
+
+    Parameters
+    ----------
+    smoothing : float, default=0.1
+        Smoothing factor for label smoothing.
     """
 
-    def __init__(self):
+    def __init__(self, smoothing=0.1):
```
rflamary, line 383: non 0 default? it that used in the paper?

antoinecollas: they use 0 but I guess we want something robust and stable by default?

antoinecollas: log(1-prob) is likely to never work

tgnassou (approved):

PR #290, opened 2025-01-10
[DRAFT] Confounded domain adaptation

calvinmccarter:


rflamary:
Hello @calvinmccarter ,

Thanks for the PR. I know this is a draft but a few quick comment before we do a proper review.

+ You need to use the SKADA API: we do not implement a fit with Xs, Ys, Xt, Yt but with X,y, sample_domain where X and y are concatenated and sample_domain contains the domain label (see other Adapter classes in SKADA. Your class must be able to be used in a DA pipeline seamlessly. 
+ I am OK with providing target labels (which is required when the counfoundng variable is the target label) but it seems to me that it could be another variable and you should allow your method to do both. So instead of providing Y I would add another variable called   `sample_confounder` (can it be multivariate?) that is used for fitting the mapping instead of using Y. If you implement only with Y as confoundrer then you clearly need to state in the class Name that this is a supervised DA method (as opposed to all other unsupervised methods in SKADA) because it would be a source of much confusion. Maybe we need to create a new Class Called SupervisedAdapter in this case.
+ the miceforest dependency need to be fully optional that is the code should run without it installed and raise an error if your class is used an you need it.

Best,

PR #292, opened 2025-02-19
[MRG] Fix Github action version v2 -> v4 and fix test with skorch update

tgnassou:
Change the version of GitHub action from v2 to v4.

Fix the test with DEV and deep method since the skorch updates change the type of `NeuralNet`.


tgnassou:
The codecov is wrong. I tested locally and all is fine!

PR #305, opened 2025-06-10
[WIP] Test time deep method

mbarneche:


skada/deep/_test_time.py
```diff
@@ -0,0 +1,128 @@
...
+                sample_idx_s=sample_idx_s,
+                sample_idx_t=sample_idx_t,
+            )
+
+
+class TestTimeNet(DomainAwareNet):
+    def __init__(self, module, criterion: "TestTimeCriterion", **kwargs):
+        super().__init__(module, criterion=criterion, **kwargs)
+
+    def fit_prepare(
```
tgnassou, line 107: I'm not fan of fit_prepare but I don't have other chouchou for now

skada/deep/_test_time.py
```diff
@@ -0,0 +1,128 @@
...
+            needed to compute DA loss:
+                - y_pred : prediction of the source and target domains
+                - domain_pred : prediction of domain classifier if given
+                - features :  features of the chosen layer
+                  of source and target domains
+                - sample_domain : giving the domain of each samples
+        y_true :
+            The true labels. Available for source, masked for target.
+        """
+        y_pred, domain_pred, features, sample_domain, sample_idx = y_pred
```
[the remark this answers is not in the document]
mbarneche: If I understand correctly, there is no need for all the `if _thing_ is not None` clauses. Is that correct?
Does this also mean that the train_on_target attribute has no use? Because using target data only depends on the mode now, right?

skada/deep/_test_time.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class TentNet(TestTimeNet):
+    def __init__(self, module, criterion, **kwargs):
+        super().__init__(module, criterion=criterion, **kwargs)
+
+    def forward(self, X, sample_domain=None, sample_weight=None):
+        """Forward pass of the model."""
+        X = self._prepare_input(X, None, sample_domain, sample_weight)
+        # Note: we are supposed to freeze some attributes of the model
```
tgnassou, line 112: I think we should be able to choose what to freeze in the TestTimeNet. We don't need a TentNet I think

skada/deep/_test_time.py
```diff
@@ -0,0 +1,155 @@
...
+        # during the adaptation phase, but we do not do it here.
+        return self.module(X)
+
+
+class TentLoss(BaseDALoss):
```
tgnassou, line 117: don't forget the docstring

skada/deep/_test_time.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(
+        self, reg_dist=1, reg_cl=1, base_criterion=None, target_criterion=None
+    ):
+        super().__init__()
+        self.reg_dist = reg_dist
+        self.reg_cl = reg_cl
+        self.base_criterion = base_criterion
+        self.target_criterion = target_criterion
+
+    def forward(self, y_s, y_t):
```
tgnassou, line 127: what is y_s and y_t ? You should have only y_t no ?

Issue #245, opened 2024-09-23
 ValueError in `CircularValidation` Scorer with Deep Models


YanisLalou:
#### Description

When using the `CircularValidation` scorer with deep models, a `ValueError` is raised during the inverse transformation of predicted source labels. This error occurs because the predicted labels (`y_pred_source`) contain previously unseen labels that are not present in the original labels (`y`).

#### Traceback

```
Traceback (most recent call last):
  File "/mnt/beegfs/workdir/yanis.lalou/anaconda3/envs/env_8_7_24/lib/python3.10/site-packages/sklearn/metrics/_scorer.py", line 141, in __call__
    score = scorer(estimator, *args, **routed_params.get(name).score)
  File "/mnt/beegfs/workdir/yanis.lalou/anaconda3/envs/env_8_7_24/lib/python3.10/site-packages/skada/metrics.py", line 44, in __call__
    return self._score(estimator, X, y, sample_domain=sample_domain, **params)
  File "/mnt/beegfs/workdir/yanis.lalou/anaconda3/envs/env_8_7_24/lib/python3.10/site-packages/skada/metrics.py", line 649, in _score
    y_pred_source = le.inverse_transform(y_pred_source)
  File "/mnt/beegfs/workdir/yanis.lalou/anaconda3/envs/env_8_7_24/lib/python3.10/site-packages/sklearn/preprocessing/_label.py", line 160, in inverse_transform
    raise ValueError("y contains previously unseen labels: %s" % str(diff))
ValueError: y contains previously unseen labels: [63]
```

#### Dataset/Method to Reproduce

**Dataset:** OfficeHome
**Method:** Deep_NO_DA_SOURCE_ONLY

#### Additional Context

This issue is likely to occur with deep models where the model can output labels that are not necessarily present in the training data.

PR #311, opened 2025-06-19
 Fix CircularValidation scorer crash on unseen predicted labels and add a test (issue #245)

arthurdrk:
This PR addresses issue #245 where the CircularValidation scorer would raise a ValueError when the backward estimator predicts labels that were not seen during training.

Changes:

The scorer now ignores predictions with unseen labels when computing the score, instead of raising an exception.
Added a dedicated test, `test_circular_validation_unseen_labels`, to ensure that the scorer handles unseen labels and does not crash.


skada/metrics.py
```diff
@@ -631,7 +631,13 @@ def _score(self, estimator, X, y, sample_domain=None, **params):
...
             if y_type == Y_Type.DISCRETE:
                 # We go back to the original labels
-                y_pred_source = le.inverse_transform(y_pred_source)
+                try:
+                    y_pred_source = le.inverse_transform(y_pred_source)
+                except ValueError as e:
+                    raise ValueError(
+                        "CircularValidation: predicted source labels contain "
+                        "unknown labels not seen during label encoding."
+                    ) from e
```
antoinecollas, line 640: I prefer that you directly check the issue https://github.com/scikit-adaptation/skada/issues/245 with a `if ...`. The code right now might print
```
"CircularValidation: predicted source labels contain unknown labels not seen during label encoding."
```
for a totally unrelated problem.

skada/tests/test_scorer.py
```diff
@@ -232,6 +232,52 @@ def test_circular_validation(da_dataset):
     assert ~np.isnan(score), "the score is computed"
 
 
+def test_circular_validation_unseen_labels():
```
antoinecollas, line 235: You should just test that you raise an error when #245 happens.

For example,
```python
# Patch the backward estimator to predict an unseen label on the sources
....
```
seems unrelated.

PR #328, opened 2025-06-24
PR update CI for testing

lionelkusch:


rflamary:
Hello, not a fan of changing the names of the actions :  Main_workflow / Linter / Lint (pull_request) instead of Lint.
Let su keep them simple if possible witrh the same naems as before and just the additiional weekly Test-datasets

PR #333, opened 2025-06-25
[TO REVIEW] Update CONTRIBUTING.md

ambroiseodt:
Update the dependency installation in CONTRIBUTING.md to solve the import problem for users using .venv virtual env and not conda virtual envs.

*Update*: the problem was coming from using zsh or bash terminal and not on the kind of virtual env used

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ GitHub, clone, and develop on a branch. Steps:
    ```bash
    $ pip install -e .
    ```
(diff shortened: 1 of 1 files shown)


antoinecollas:
Is it related to venv vs conda or is it because of your terminal?
In some terminals, you need to escape brackets: `pip install -e .\[dependance\]`

ambroiseodt:
After some tests, I confirm that it comes from zsh. Adding the quotes works fine for any terminal so I think you can merge. Thanks

PR #337, opened 2025-06-25
Handle torch tensor in DomainAwareDataset

tom-yneuro:
Adding support for flexible data types (`numpy` arrays and `torch` tensors) and a new return type (`DeepDADataset`) inside `DomainAwareDataset`. 

### Enhanced `pack` Method Functionality:
* Implemented logic in the `pack` method to handle data concatenation and return the appropriate format based on the `return_type`. Added checks for `torch` availability and raised appropriate errors if required dependencies are missing.
* Added support for returning data as a `DeepDADataset` object, which is compatible with PyTorch's `Dataset` interface and includes domain-specific information.

### Testing Enhancements:
* Added a new test in `test_deep_dadataset.py` to validate the functionality of the `pack` method with different `return_type` values (`"array"`, `"tensor"`, `"DeepDADataset"`). Ensures that the output matches the expected types and formats.


tom-yneuro:
I noticed that the behavior of `return_X_y` isn't obvious when set to False (i.e it returns a `Bunch` object or a `DomainAwareDataset` depending on the file) 
Here I proposed to change this argument to `return_type`, in my case allowing several types 
Maybe this approach can be generalized to datasets fetch with `return_X_y` argument

tgnassou (approved):

tgnassou:
LGTM!

PR #343, opened 2025-07-02
Use Flag for improving the management of CodeCov

lionelkusch:
Based on the [flags](https://docs.codecov.com/docs/flags) documentation of CodeCov.
I try to improve the management of the coverage.

I will try next week to test it and see if it's working or not. 

This is linked to the issue of the coverage of some Pull Requests are not passing because the dataset tests are not running.


lionelkusch:
I created Proof of concept: [here](https://github.com/lionelkusch/skada_2)
I mostly fix everything. The only missing point is that the `dataset` flag considers all the files. I don't understand why. 

@rflamary
Looking more into details of the tests in the dataset folder, I remarked that these tests don't test all the functions in the folder and some tests outside of this folder test some specific functions of it.

If I limit the tests in the dataset folder, I will reduce the coverage of the folder by 1.7 % of the dataset folder.
The other solution is to run all the tests for keeping the same coverage but I don't really like this solution.

What do you think?

PR #344, opened 2025-07-02
WIP: add Alpha parameter for relaxed DA 

mgeneralince:


examples/methods/plot_optimal_transport_da.py
```diff
@@ -395,3 +397,97 @@
 plt.scatter(X_target[:, 0], X_target[:, 1], c=y_target, vmax=9, cmap="tab10", alpha=0.7)
 plt.axis(lims)
 plt.title(label=f"OTDA linear (ACC={ACC_linear:.2f})")
+
+# %%
+# Partial mapping with alpha parameter
+# ------------------------------
+# The OTDA method can be used with a parameter alpha that controls the amount of
+# transport applied to the source samples following the expression
+# X_mapped = (1-alpha)*X_target + alpha*OT(X_target).
```
tgnassou, line 406: It should be X_source here no ?

examples/methods/plot_optimal_transport_da.py
```diff
@@ -395,3 +397,97 @@
...
+    plt.title(f"Alpha = {alpha:.2f} | ACC(target) = {acc_target:.2f}")
+    plt.legend()
+
+    return 1
+
+
+ani = animation.FuncAnimation(
+    plt.gcf(), _update_plot, len(alphas), interval=200, repeat_delay=2000
+)
+ani.save("otda_animation_1.gif")
```
tgnassou, line 493: @rflamary wdyt ? We keep the gif animation for the docs? I dont think you can print a gif ?

skada/_mapping.py
```diff
@@ -100,6 +106,7 @@ def transform(
         X_adapt, _ = source_target_merge(
             X_source, X_target, sample_domain=sample_domain
         )
+        X_adapt = self.alpha * X_adapt + (1 - self.alpha) * X
```
tgnassou, line 109: The alpha transformation should be done on X source not on X_adapt ? I think it is the same but right now for target you are doing: alpha * X_target + (1 -alpha) * X_target =X_target

skada/_mapping.py
```diff
@@ -141,11 +148,13 @@ def __init__(
         metric="sqeuclidean",
         norm=None,
         max_iter=100_000,
+        alpha=1.0,
```
tgnassou, line 151: Alpha is missing in the docstrings

PR #345, opened 2025-07-10
[MRG] Test time API for shallow method

tgnassou:
The goal of this PR is to propose a API for test time method. 

the test time method will break if the new domain has never be seen but the model, and will raise a error to update the model with `fit_new_domain`. To work with pipeline, one can enable `auto_fit_new_domain`. With this on, it will automatically refit for new unseen domain.

@rflamary if you want to check!


skada/base.py
```diff
@@ -113,6 +113,20 @@ def fit(self, X, y=None, *, sample_domain=None, **params):
...
 
+class BaseTestTimeAdapter(BaseAdapter):
+    """Base class for test-time adapters.
+
+    Test-time adapters are used to adapt the data during evaluation, e.g. to
+    align the target domain with the source domain. The main difference with
+    the `BaseAdapter` is that test-time adapters do not require fitting and
+    can be used directly in the `transform` method.
+    """
+
```
rflamary, line 127: should you also define a new init function with fit_new_domain arable or is it done independently by individual implementations? probably better not to do that since base does not either

skada/base.py
```diff
@@ -113,6 +113,20 @@ def fit(self, X, y=None, *, sample_domain=None, **params):
...
+    def fit_new_domain(self, X, y=None, *, sample_domain=None, **params):
+        """Fit the adapter to the new domain."""
+        pass
```
rflamary, line 132: raise notimplented like for BaseAdapter?

rflamary (commented):
LGTM, a few comments below but nothing too important

rflamary:
> @rflamary Maybe it is the time to revive this PR ? WDYT ?

Oly took me 6 months ;). Sorry about that

